PR #2, opened 2016-09-06
Add Storage module for localStorage and sessionStorage support

vramana:


chenglou:
Cool! Thanks but I have a few thoughts on this.

Since localStorage and sessionStorage are singletons, I think we should be consistent and type them the same way we type window. This way, we'd do `LocalStorage.setItem` rather than `setItem localStorage`

Also, I don't think there's guarantee that LocalStorage and SessionStorage will have the same API in JS. We could probably duplicate their definitions (not too much trouble right now anyway).

chenglou:
Thanks!

chenglou:
API is still in flux btw. So we might revert to your API later on ¯\_(ツ)_/¯

vramana:
Haha. Okay.

src/index.re
```diff
@@ -12,3 +12,21 @@ let module Window = {
...
+  let getLength () => [%bs.raw {| localStorage.length |}];
+};
+
+let module SessionStorage = {
+  external getItem : string => 'a = "sessionStorage.getItem" [@@bs.val];
+  external setItem : string => 'a => unit = "sessionStorage.setItem" [@@bs.val];
+  external removeItem : string => unit = "sessionStorage.removeItem" [@@bs.val];
+  external clear : unit => unit = "sessionStorage.clear" [@@bs.val];
+  external key : int => 'a = "sessionStorage.key" [@@bs.val];
+  let getLength () => [%bs.raw {| sessionStorage.length |}];
```
chenglou: Hmm good point. We can type it as `  external length : int = "localStorage.length" [@@bs.val];` but the compiler might inline the value and produce the wrong output since `length` is mutative. @bobzhang what'd be the solution here? Maybe we need to switch back to using plain functions without module?

bobzhang: below is my proposal

```ocaml
type sessionStorage 
external sessionStorage : sessionStorage = "" [@@bs.val]
external getItem : sessionStorage -> string -> 'a = "" [@@bs.send]
external length : sessionStorage -> int = "" [@@bs.send]
```
The same applies to local storage too

chenglou: Hah, we're back to the original PR then!

vramana: 😀 I'll submit it in my other PR then.

PR #5, opened 2016-09-19
Make ReasonJs rebel compatible + additions

bsansouci:


bsansouci:
Killing in favor of a new PR.

PR #8, opened 2016-12-19
Some functions required for WebGL

evilpie:
This is probably not perfect we just started learning Reason/Ocaml, but those functions are useful for setting up a minimal WebGL example. We would appreciate your feedback.

### DIFF
--- a/src/reasonJs.re
+++ b/src/reasonJs.re
@@ -4,6 +4,7 @@ module Window = {
   external innerWidth : t => int = "innerWidth" [@@bs.get];
   external innerHeight : t => int = "innerHeight" [@@bs.get];
   external addEventListener : t => string => (unit => unit) => unit = "addEventListener" [@@bs.send];
+  external requestAnimationFrame : t => (unit => unit) => unit = "requestAnimationFrame" [@@bs.send];
   external onLoad : t => (unit => unit) => unit = "onload" [@@bs.set];
 };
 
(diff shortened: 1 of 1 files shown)


bsansouci:
@evilpie Hey! Nice PR. I don't know what @chenglou thinks about it, but in the meantime you can checkout [regl](https://github.com/bsansouci/regl) which wraps all of those for both OpenGL and WebGL :D

src/reasonJs.re
```diff
@@ -4,6 +4,7 @@ module Window = {
   external innerWidth : t => int = "innerWidth" [@@bs.get];
   external innerHeight : t => int = "innerHeight" [@@bs.get];
   external addEventListener : t => string => (unit => unit) => unit = "addEventListener" [@@bs.send];
+  external requestAnimationFrame : t => (unit => unit) => unit = "requestAnimationFrame" [@@bs.send];
```
chenglou: Let's put this outside, standalone like setTimeout. This is available on server-side without window.

src/reasonJs.re
```diff
@@ -22,12 +23,145 @@ module Math = {
...
 module Document = {
   type element;
   external getElementById : string => element = "document.getElementById" [@@bs.val];
   external appendChild : element => 'element => unit = "appendChild" [@@bs.send];
+
+  /* Should be on CanvasElement */
+  external getContext : element => string => GL.glT = "getContext" [@@bs.send];
 };
 
+external log: string => unit = "console.log" [@@bs.val];
```
chenglou: No need for this! `Js.log` and `print_endline` work =)

chenglou:
Since there aren't big changes needed, I'll merge this and make the two minor corrections. Thanks!

chenglou:
Btw @evilpie: just ran this and BS caught an `external` bug: `vertexAttribPointer` should be passed a `Js.boolean`, not a bool. I'm fixing this.

evilpie:
@chenglou Thanks for merging, I almost forgot about this. I think in the future it would be wise to generate the bindings directly from the WebIDL files.

PR #11, opened 2017-02-04
Fleshes out the Date module and a bit of the DOM API

glennsl:
__NOTE: This PR has been heavily expanded post-submission.__

This mostly completes the following modules:
* `Date`
* `Document`
* `Element` down to `HTMLElement`
* `History`
* `Location`
* `Range`
* `Selection`

And partly implements many more.

It breaks backwards compatibility with (at least) the following:
* `Window.window`
* `Window.addEventListener`
* `Window..Location.href`
* `Document.getElementById`
* `Document.getElementsByClassName`
* `Promise` functions rhave been enamed and converted to pipe form
* functions previously in `Document` have been moved to more appropriate modules.
* All Dom bindings with `@@bs.set` have been converted to pipe form. This does not include setters and other modules like `GL`.
* Pretty much everything that used `Js.null` or `Js.boolean` (they're now transformed into option and bool respectively)

Open questions:
* How to organize things? Where should enums and constants go and how should they be structured? Where should forward declarations go? Should variants be wrapped in a module (for namespacing and bundling of related logic)?
* Should all types be forward declared, just in case?
* Should enum variants fail on encode or decode? Or should enums have an Unknown variant hat encodes to some nonsense that will break something later (current approach)?
* What should the convention for constants be? Currently `_MyConstant`.
* Should "down-casting" be done with polymorphic or regular variants?
* Should setters manually be converted to pipe form?
* Are the current API conventions good enough to commit to? Should some or all of it be considered and documented as unstable?
* What are the guidelines on what ought to be included here? And what should be implemented in the BuckleScript std lib instead? E.g. `String` and `Array` are (partially) implemented in BuckleScript.
* What should the policy be for including experimental, deprecated, and "de-facto standard" APIs?
* It's not always clear where dom APIs should go. E.g. `CanvasElement` which is a dom element but also exposes webgl functionality. Should it be put in dom, and the dom depend on `GL`? Is that future-proof?
* What should the convention for overloads be? The current convention of "*With" followed by a list of the parameter names included is a bit verbose (lol at `Date`)
* What shouold the naming convention for abbreviations be? Due to OCaml's rules for which identifiers can be lower case and upper case, the current convention is PascalCase/camelCase, e.g. "Html", "Json", "dom" etc.
* Lack of inheritance leads to lots of casting and is a bit cumbersome. Can it be done better? Including a functor maybe?

Todo:
* `Event` type-specific functions should be pulled out of `Event`
* `HtmlElement` type-specific functions should be pulled out of `HtmlElement`
* Lots of the dom still missing
* Lots of other functionality still missing?
* Functional tests, not just "cosmetic"

---

**Initial PR text:**

Based roughly on MDN for reference:
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Date

Notes:
- Gecko-specific methods were skipped
- Overloads have been given a somewhat ridiculous naming scheme
- `makeWithString` overload skipped for the equivalent and better named `parse` function
- Some deprecated methods have been included, but marked as deprecated
- Overloads for the `toLocale*` methods have been skipped because they're pretty complicated, but they're also not widely supported, so I'm sure it's fine to leave them for now.


glennsl:
@chenglou Ok, module names have been mangled and aliased on ReasonJs. I've also updated the PR description. Any feedback, on the open questions or otherwise, would be greatly appreciated.

Do you need me to chill out a bit so you can go through this, or can I continue to expand the API?

glennsl:
See https://github.com/glennsl/rehydrate/commit/ac7b1b8561fc1b71ef80f3012202c33cd79690da for an example of the changes needed after merging this PR.

(The API would be nicer to use if there was a stdlib `Option` module with `map` and such)

chenglou:
Hot fucking damn, I think I should merge this rather than squash it, to make your contribution show adequately. Thanks so much!

glennsl:
I don't even know what squashing is, but thanks! :)

PR #14, opened 2017-02-10
Some fixes and additions

glennsl:
* Fleshed out `Promise`. Changed the signature of `Promise.make`, otherwise just additions.
* Removed `Style` module, because it's not actually a thing
* Added `CssStyeDeclaration` which is the actual thing that `Style` thought it was.
* Added a few down-casting functions `Document.ofNode`, `Element.ofNode` and `HtmlElement.ofElement`
* Removed some semicolons from @@bs.raw statements that weren't agreeing with webpack.

### DIFF
--- a/src/core/promiseRe.re
+++ b/src/core/promiseRe.re
@@ -1,8 +1,15 @@
 type t 'a = CoreRe.promise 'a;
 
+/* TODO: Include error in type? */
+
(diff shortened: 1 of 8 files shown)


chenglou:
Thanks! Yeah there's a difference between `[%bs.raw]` and `[%%bs.raw]`

PR #15, opened 2017-02-11
Added implementation inheritance

glennsl:
It's not terribly elegant, but it works and means less casting is needed. Specifically:
* Implementations are inherited down the tree. This means e.g. `Node.innerText` now also exists as `Element.innerText` for use on `Element`s
* There is no sub-type polymorphism, so types will still need to be manually up-casted when used with methods that expect a specific type. E.g. an `Element` will need to be casted to a `Node` to be appended with `Node.appendChild`.
* Down-casting is still needed even when you know exactly what the element is, e.g. `Document.createElement "div"` will return an `Element`, not an `HTMLDivElement`. I have some ideas on how to this might be solved too though.

Another consequence is that the `type t` convention can't be used as each included module needs to contain unique names. `DomRe` has therefore been exposed as `ReasonJs.Dom` so that e.g. what was `ReasonJs.Element.t` can now be accessed as `ReasonJs.Dom.element`.

### Implementation details:

Consider this module
```reason
type t = fooType;

let foo : t => t = ...;

/* also want to include functions of some "parent" type */
```

The pattern employed wraps "inheritable" modules in a functor that by convention should be named `Impl`:
```reason
module type TypeSig = { type t };

module Impl (Type: TypeSig) => {
  type t_foo = Type.t; /* (1) */
  let foo : t_foo => fooType = ... /* (2) */
};

include Parent.Impl { type t = fooType };
include Impl { type t = fooType };
```

et voilà!

###### Notes:
1. `type t` needs to be renamed to avoid name clashes with other inherited modules, like `Parent.Impl` in this example
2.  In most cases arguments and return types other than "self" should not be `t_foo` since, when inherited, t_foo will take on the type passed to it.

### DIFF
--- a/examples/dom_example.re
+++ b/examples/dom_example.re
@@ -2,12 +2,75 @@ open ReasonJs;
 
 /* Adapted from https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Examples#Example_7:_Displaying_Event_Object_Properties */
 
-open Document;
-open Element;
+
+/* Option.map */
+let map f => fun
+| Some v => Some (f v)
+| None => None;
+
+let and_then f => fun
+| Some v => f v
+| None => None;
+
+let unwrapUnsafely = fun
+| Some v => v
+| None => assert false;
+
+
+/*
+ * These SHOULD type check
+ */
+document
+  |> Document.createElement "div"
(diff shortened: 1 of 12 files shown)


glennsl:
Merged with #23

PR #17, opened 2017-02-12
fix the test js files

dorafmon:


chenglou:
Thanks! But let's just disable the header completely. See my comment in the above commit.

cc @bobzhang. I'm not too sure `-no-bs-version-header` should be a bsc/bsb default.

PR #34, opened 2017-02-26
Replaced Date and Promise with Js.Date and bs-promise

glennsl:
cc @chenglou Can you review? This has breaking changes.

(Replaces #22 if merged)


chenglou:
Accidentally merged (but reverted) #22. This should be ok?

glennsl:
oops! Yea, looks good!

PR #37, opened 2017-02-27
Adds Travis CI

glennsl:
cc @chenglou For approval. I've requested org access from travis CI as well


glennsl:
Got access to org in travis, but apparently I need admin access to enable travis for the actual repo.

chenglou:
Made you admin

PR #45, opened 2017-03-20
Moves all enum variants and encode/decode functions to DomTypesRe

glennsl:
cc @chenglou This is technically a breaking change. I could alias the types back and deprecate them, but I don't think it should cause much trouble. What do you think?


chenglou:
Nothing public's depending on master.

chenglou:
DomTypesRe is still included in the bspacked bundle though?

glennsl:
Aren't you the one doing the packing? I would assume it is, but I don't know much about the process.

chenglou:
Well, domRe depends on domTypesRe, so I have to bundle these too.

glennsl:
I don't have the `.ml` files to run it against. How do you generate those?

chenglou:
I've put them in https://github.com/BuckleTypes/reason-js/tree/test-bspack

glennsl:
Yea, but those are obsolete with the changes made in this PR, so they would have to be regenerated.

glennsl:
Fucking discord... Are you ok to merge this? Think we should be more aggressive with size reduction?

chenglou:
We should be more aggressive, yes. I'm willing to compromise completeness (heck, maybe even soundness, if we provide a few unsafe casts) to reduce learning overhead too. I'd like the newcomer's impression when manipulating the DOM to be "hey this is almost the same, but typed!" instead of "where can I find getElementById" lol

PR #46, opened 2017-03-20
Fixes Fetch.RequestInit.make defaults

### LINKED ISSUE #44: Invalid referrer policy when using `fetchWithInit`
When using the [fetchWithInit example](https://github.com/BuckleTypes/reason-js/blob/master/examples/fetch_example.re#L11), I get an error from the browser: `TypeError: Failed to execute 'fetch' on 'Window': Invalid referrer policy`

According to the [spec](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy), referrerPolicy should never be undefined. 

Adding None as the referrerPolicy works:

```
Fetch.RequestInit.make method_::Post referrerPolicy::None ();
```

glennsl:
Hopefully fixes #44. Waiting for verification

### DIFF
--- a/src/fetch/fetchRe.re
+++ b/src/fetch/fetchRe.re
@@ -265,49 +265,44 @@ module RequestInit = {
   let map f => fun /* internal */
   | Some v => Some (f v)
   | None => None;
-  /*
   external make :
-    method_::requestMethod? =>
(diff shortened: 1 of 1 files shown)


glennsl:
Hoping for the best...

PR #49, opened 2017-03-28
Fix FFI which return options

mransan:
When retuning option from an `external` the attribute [@@bs.return null_undefined_to_opt] must be added.


glennsl:
Thanks!

PR #50, opened 2017-03-29
domRect getters

ccorcos:
@glennsl 

reference: https://developer.mozilla.org/en-US/docs/Web/API/DOMRect


glennsl:
Looks good! Could you also add [the constructor](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect/DOMRect) as `make` (similar to [this](https://github.com/BuckleTypes/reason-js/blob/master/src/dom/ranges/rangeRe.re#L3) but with more arguments), and a module alias in `DomRe.re`?

ccorcos:
done

glennsl:
Sorry for not clarifying this before. Could you add the module to [the list of modules here](https://github.com/BuckleTypes/reason-js/blob/master/src/dom/domRe.re)?

src/dom/geometry/domRectRe.re
```diff
@@ -1,15 +1,19 @@
-external top : Dom.domRect => int = "" [@@bs.get];
+type t = Dom.domRect;
 
-external bottom : Dom.domRect => int = "" [@@bs.get];
+external make : int => int => int => int => t = "DOMRect" [@@bs.new]; /* experimental */
```
glennsl: Could you also add labels to these arguments? With four arguments of the same type it's very easy to mistake them.

ccorcos:
alrighty, let me know

glennsl:
Thanks! I had to remove some text that had been randomly inserted, but otherwise it's good!

PR #53, opened 2017-04-08
Fix README example to use Js.Global.setInterval

jasim:


glennsl:
Oh man, thanks for the heads up! But we should really replace it with an example that's actually releveant. Would you like to give that a shot? Otherwise I'll close this and conjure up something myself.

jasim:
Thanks for the quick feedback! I spent about an hour trying to figure out BuckleScript FFI and got stuck here because the example wasn't working. Made the PR because I thought other newcomers could also get similarly tripped up. 

Happy to try something a bit more involved. a) 60s stopwatch that mutates the DOM, or b) a stripped-down DOM-based TodoList. Or please go ahead if you're looking for something different!

glennsl:
Yeah, sorry about that. We've been quite aggressively moving things out of reason-js lately, into the BuckleScript std lib or separate libraries.

A TodoList is probably a bit too involved. Ideally the example would be less than 10 lines or so. Small enough to grasp at a glance. A 60s stopwatch seems like a really good example though. It'd really be appreciated!

glennsl:
I'm merging this, since it continues to trip people up. Thanks for this, and feel free to open a new PR with a better example.

Issue #55, opened 2017-04-21
2d canvas API?


af:
Hi,
I'm writing a simple 2d canvas game, and have been building up a bunch of bindings for the CanvasRenderingContext2D API. Would it be helpful to add them in a pull request here? If so, a couple of questions:

* There's an existing `ReasonJs.CanvasElement.getContext` function that takes a string, and returns a context. However the existing context API here is for WebGL, not the 2d API. So I'd propose changing this to `getWebGLContext`, and creating a new `get2dContext` function. It looks like `[@bs.as]` will work to pass the appropriate string parameter in to js's `getContext`

* ints or floats for canvas methods? I've been using floats myself but I'm wondering if there's a case for integer arguments

glennsl:
Hi! I think that'd be a yes on all counts. Except perhaps the last one, where a "yes" wouldn't really make sense. There I guess I'd say to stick with floats until it's evident that ints would be much more convenient. Floats are pretty safe.

glennsl:
Closing this since #56 has been merged, though it's not complete it's a very good start!

PR #56, opened 2017-04-25
Add 2d canvas bindings (for #55)

af:
This isn't a complete set of bindings, but I believe all remaining context methods are marked as TODO. I'll try and circle back to add most of the rest but might not get around to it for a bit.

TODO:

- [x] Convert to bs.send.pipe
- [x] Use labelled args where needed
- [x] Add tests


src/gl/canvas2d.re
```diff
@@ -0,0 +1,76 @@
+type ctx;
```
glennsl: It's conventional to name the primary type of a module just `t`.

src/gl/canvas2d.re
```diff
@@ -0,0 +1,76 @@
...
+
+/* Canvas API, following https://simon.html5.org/dump/html5-canvas-cheat-sheet.html */
+
+external save : ctx => unit = "save" [@@bs.send];
```
glennsl: You don't have to specify `"save"` since it will be inferred for you based on the name of the `external`. you can just leave it as an empty string: `external save : ctx => unit = "" [@@bs.send];`

src/gl/canvas2d.re
```diff
@@ -0,0 +1,76 @@
...
+external restore : ctx => unit = "restore" [@@bs.send];
+
+/* Transformation */
+external scale : ctx => float => float => unit = "scale" [@@bs.send];
```
glennsl: Could you convert the `send` externals to `send.pipe`? E.g. `external scale : float => float => unit = "" [@@bs.send.pipe: ctx];`. It will put the `ctx` argument at the end instead, so you can then directly use it with the pipe operator (`|>`) and more easily apply it partially. This is the more idiomatic form.

src/reasonJs.re
```diff
@@ -2,11 +2,13 @@ external requestAnimationFrame : (unit => unit) => unit = "" [@@bs.val];
 
 module Base64 = Base64Re;
 module Gl = GlRe;
+module Canvas2d = Canvas2d;
```
glennsl: Can you name the module `Canvas2dRe`? This is to keep internal modules from conflicting with modules defined in its dependents. You should still keep the alias as it is, though, so it won't affect how dependents actually consume it.

src/reasonJs.re
```diff
@@ -2,11 +2,13 @@ external requestAnimationFrame : (unit => unit) => unit = "" [@@bs.val];
...
 
 module Storage = StorageRe;
 
 module CanvasElement = {
-  external getContext : Dom.element => string => Gl.glT = "" [@@bs.send];
+  external getWebglContext : Dom.element => (_ [@bs.as "webgl"]) => Gl.glT = "getContext" [@@bs.send];
+  external get2dContext : Dom.element => (_ [@bs.as "2d"]) => Canvas2d.ctx = "getContext" [@@bs.send];
```
glennsl: I'm wondering if these shouldn't be `getContext2d` and `getContextWebGl` to aid in discoverability.

af:
@glennsl Thanks for the feedback! I'm new to OCaml/bucklescript so it's really helpful to learn more of the conventions (and I wasn't aware of `bs.send.pipe`). I've made most of the changes locally but now I'm wondering if the `[@@bs.set]` externals should have the context parameter at the end as well, so they can also be chained with the pipe operator?

glennsl:
@af Unfortunately there's no `set.pipe` yet, though I have requested it (and made an early attempt at implementing it, but got tripped up with build issues)

af:
@glennsl Ah I see, so just putting the context as the last parameter won't make it automatically chainable I guess? Bucklescript needs to have the external "opt-in" to the piping?

glennsl:
The `@@bs` attributes specify how BuckleScript will interpret the type signature. `bs.set` and `bs.send` interpret the first argument as `this`. So with `external f : ... [@@bs.send]`, `f x y` will generate `x.f(y)`, and with `external f : ... [@@bs.set]` it will generate `x.f = y`.

af:
@glennsl rebased on top of newest master and made your suggested changes (still need to add tests, will do that in the next day or two). I'm trying to use this branch in the game I'm writing, but I have an issue when trying to compile this code:

```
  ctx
    |> C.translate x y
    |> C.rotate theta;
```

The bucklescript compiler complains about the `C.rotate` line (I commented out the rest of my code for simplicity): 

```Error: This expression has type C.t -> unit
       but an expression was expected of type unit -> 'a
       Type C.t = Canvas2dRe.t is not compatible with type unit
```

Any pointers for that? I can't see the issue from looking at the bindings

chenglou:
I'm thinking we should probably pull this out into a dedicated bs-canvas repo. Also makes it API discovery easier. Our one-file-bindings repos are all pretty simple and clean.

cc @bassjacob

bassjacob:
@chenglou I could go either way on it being a separate repo. Wouldn't this repo need to depend on bs-webapi for the Dom and Gl types?

chenglou:
I don't see any dependency in this PR on bs-webapi's types?

glennsl:
@af They'll have to return `t` instead of `unit` to be able to chain them like that (the second `|>` will use the result of evaluating `ctx |> C.translate x y` as the last argument to `C.rotate`). Unfortunately the underlying JS API doesn't, so this isn't going to be possible unless you wrap the externals (which isn't recommended because it'll add overhead).

@chenglou Not sure `canvas` is the right "granularity". I'm no expert on this part of the APIs, which is why I've been on the fence, but I imagine you'll usually need `requestAnimationFrame` for example, so you'll still need to discover anbd depend on `bs-webapi`. Perhaps there should be a separate package that includes rAF and other related APIs, but then what would that be called? `bs-canvas` would not be a good name for people looking for just rAF.

@bassjacob `Gl` could be moved to `canvas` I think, and `Dom` is shipped with BuckleScript.

chenglou:
requestAnimationFrame/cancelAnimationFrame should probably reside in its own repo with a `[%external __DEV__]` to detect whether it's web or node?

src/gl/canvas2dRe.re
```diff
@@ -0,0 +1,76 @@
...
+
+external save : unit = "" [@@bs.send.pipe: t];
+external restore : unit = "" [@@bs.send.pipe: t];
+
+/* Transformation */
+external scale : float => float => unit = "" [@@bs.send.pipe: t];
+external rotate : float => unit = "" [@@bs.send.pipe: t];
+external translate : float => float => unit = "" [@@bs.send.pipe: t];
+external transform : float => float => float => float => float => float => unit = "" [@@bs.send.pipe: t];
+external setTransform : float => float => float => float => float => float => unit = "" [@@bs.send.pipe: t];
```
glennsl: This looks like a prime use-case for labeled arguments. But the argument names proposed by MDN doesn't seem terribly helpful either:
```
external transform : m11::float => m12::float => m21::float => m22::float => dx::float => dy::float => unit = "" [@@bs.send.pipe: t];
```

src/gl/canvas2dRe.re
```diff
@@ -0,0 +1,76 @@
...
+external stroke : unit = "" [@@bs.send.pipe: t];
+external clip : unit = "" [@@bs.send.pipe: t];
+external moveTo : float => float => unit = "" [@@bs.send.pipe: t];
+external lineTo : float => float => unit = "" [@@bs.send.pipe: t];
+external quadraticCurveTo : float => float => float => float => unit = "" [@@bs.send.pipe: t];
+external bezierCurveTo : float => float => float => float => float => float => unit = "" [@@bs.send.pipe: t];
+external arcTo : float => float => float => float => float => unit = "" [@@bs.send.pipe: t];
+external arc : float => float => float => float => float => bool => unit = "" [@@bs.send.pipe: t];
+external rect : float => float => float => float => unit = "" [@@bs.send.pipe: t];
+external isPointInPath : float => float => bool = "" [@@bs.send.pipe: t];
```
glennsl: More opportunities for labeled arguments here

src/gl/canvas2dRe.re
```diff
@@ -0,0 +1,76 @@
...
+
+/* Text */
+external font : t => string => unit = "" [@@bs.set];
+external textAlign : t => string => unit = "" [@@bs.set];
+external textBaseline : t => string => unit = "" [@@bs.set];
+external fillText : string => float => float => unit = "" [@@bs.send.pipe: t];
+external strokeText : string => float => float => unit = "" [@@bs.send.pipe: t];
```
glennsl: And here

src/gl/canvas2dRe.re
```diff
@@ -0,0 +1,76 @@
...
+/* TODO: measureText(); optional maxwidth arg for fillText/strokeText */
+
+/* Rectangles */
+external fillRect : float => float => float => float => unit = "" [@@bs.send.pipe: t];
+external strokeRect : float => float => float => float => unit = "" [@@bs.send.pipe: t];
+external clearRect : float => float => float => float => unit = "" [@@bs.send.pipe: t];
```
glennsl: And here

af:
@glennsl Ah, makes sense on the pipe error. I thought maybe `bs.send.pipe` might have some wizardry to wrap the call and set the return value. So should I revert to using `bs.send` for all of those externals then?

src/gl/canvas2dRe.re
af: Yeah a signature of 6 consecutive floats is pretty gross. The labelled version isn't really significantly easier to read though. I haven't used transform/setTransform directly myself, so I'm not aware of a cleaner solution

src/gl/canvas2dRe.re
af: How about using `(float, float)` tuples to represent points? Is there a way to "spread" them into a flat argument list with bs externals? (I couldn't find anything in the bs manual). Eg arcTo() would be much more readable like this:
```
external arcTo : (float, float) => (float, float) => float => unit = "" [@@bs.send.pipe: t];
```

or as labelled args:
```
external arcTo : p1::(float, float) => p2::(float, float) => r::float => unit = "" [@@bs.send.pipe: t];
```

glennsl: No, unfortunately not. I think labeled is better than unlabeled though, since it at least gives some indication of what the argument is.

glennsl:
@af I'd say don't revert. I think I still prefer them in pipe form.

af:
Re: moving to a separate package, you still need to use `Dom.element` in order to get a canvas context, so to me it makes sense to keep this alongside the DOM bindings. If/when DOM bindings are included with bucklescript, then I think it'd make sense to move the canvas API to its own repo.

chenglou:
@af We've upstreamed the types exactly for this purpose! =D

With untyped languages we either need to upstream everything or nothing. For ML typed languages, upstreaming the types strikes a nice balance between interoperability and iteration.

af:
@glennsl Added tests and made the labelled arg changes. After converting my canvas game to this API I agree it's quite a bit more ergonomic. I think this is probably good to go now.

If you guys decide to move this to a separate repo, let me know if there's anything I can do to help.

src/gl/canvas2dRe.re
```diff
@@ -0,0 +1,76 @@
+type t;
+
+/* Canvas API, following https://simon.html5.org/dump/html5-canvas-cheat-sheet.html */
+
+external save : unit = "" [@@bs.send.pipe: t];
+external restore : unit = "" [@@bs.send.pipe: t];
+
+/* Transformation */
+external scale : float => float => unit = "" [@@bs.send.pipe: t];
```
glennsl: Could you add labels (`x` and `y`) to these too?

src/gl/canvas2dRe.re
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+/* Paths */
+external beginPath : unit = "" [@@bs.send.pipe: t];
+external closePath : unit = "" [@@bs.send.pipe: t];
+external fill : unit = "" [@@bs.send.pipe: t];
+external stroke : unit = "" [@@bs.send.pipe: t];
+external clip : unit = "" [@@bs.send.pipe: t];
+external moveTo : float => float => unit = "" [@@bs.send.pipe: t];
+external lineTo : float => float => unit = "" [@@bs.send.pipe: t];
```
glennsl: And these

tests/gl/canvas2d_test.re
```diff
@@ -0,0 +1,56 @@
+
+/* TODO: make canvas tests here */
+open ReasonJs.Canvas2d;
+open ReasonJs.CanvasElement;
+open ReasonJs.Dom;
+
+/*let canvasEl = DocumentRe.querySelector "canvas" document;*/
+let canvasEl = DocumentRe.createElement "canvas" document;
```
glennsl: Just `Document.createElement` should work here, right? "Usage code" should avoid using internal modules, since they're likely to be hidden in the future.

src/gl/canvas2dRe.re
```diff
@@ -0,0 +1,76 @@
...
+/* Transformation */
+external scale : float => float => unit = "" [@@bs.send.pipe: t];
+external rotate : float => unit = "" [@@bs.send.pipe: t];
+external translate : float => float => unit = "" [@@bs.send.pipe: t];
+external transform : m11::float => m12::float => m21::float => m22::float => dx::float => dy::float => unit = "" [@@bs.send.pipe: t];
+external setTransform : m11::float => m12::float => m21::float => m22::float => dx::float => dy::float => unit = "" [@@bs.send.pipe: t];
+
+/* Compositing */
+external globalAlpha : t => float => unit = "" [@@bs.set];
+external globalCompositeOperation : t => string => unit = "" [@@bs.set];
```
glennsl: This is (kind of) an enum isn't it? Could you add a comment saying so? In the future (or now, if you're keen on it) we'll probably want to encode the enum values using a more appropriate type for added type safety.

src/gl/canvas2dRe.re
```diff
@@ -0,0 +1,76 @@
...
+
+/* Line Styles */
+external lineWidth : t => float => unit = "" [@@bs.set];
+external lineCap : t => string => unit = "" [@@bs.set];
+external lineJoin : t => string => unit = "" [@@bs.set];
```
glennsl: I think these are enums too.

src/gl/canvas2dRe.re
af: @glennsl I'm trying to figure out how to type the enums with bucklescript but keep getting stuck. The bucklescript manual uses variant constructors starting with a backtick, which Reason seems to complain about. Here's what I have as an example:

```
type compositeT =
  | SourceOver [@bs.as "source-over"]
  | SourceIn [@bs.as "source-in"]
  | SourceOut [@bs.as "source-out"]
  | SourceAtop [@bs.as "source-atop"]
  | DestinationOver [@bs.as "destination-over"]
  | DestinationIn [@bs.as "destination-in"]
  | DestinationOut [@bs.as "destination-out"]
  | DestinationAtop [@bs.as "destination-atop"]
  | Lighter [@bs.as "lighter"]
  | Copy [@bs.as "copy"]
  | Xor [@bs.as "xor"];

/* and then using it: */
```
external globalCompositeOperation : t => compositeT [@bs.string] => unit = "" [@@bs.set];
```

I'm getting `Error: Not a valid  type for [@bs.string]`. Any pointers?

Comment (glennsl) [inline src/gl/canvas2dRe.re]: There's several problems with what you're doing here. The first is that `compositeT` needs to be a polymorphic variant type (see https://realworldocaml.org/v1/en/html/variants.html#polymorphic-variants). The second problem is that the type needs to be inlined in the type annotation of the external. You can't have it as a separate `type` declaration.

This makes it a bit cumbersome and limited in use. it might be better to just have a module with constants:
```reason
module Composite = {
  type t = string;

  let sourceOver : t = "source-over";
  ...
};

external globalCompositeOperation : t => Composite.t => unit = ""
```

af: Thanks @glennsl! I fought with the polymorphic variants for a bit, and what I had compiled fine, but bucklescript was converting them to ints instead of strings somehow.

Anyways I pushed up a commit with one of the enums modeled as a sub-module. If that looks good I'll convert the rest of the string enums as well.

glennsl: @af Yea I think this looks good. A bit unfortunate that the values aren't inlined, but still better than the polymorphic variants I think. :+1:

af: Great, thanks. I added the other two enum submodules as well now. One thing I noticed though is that I can still do `lineCap ctx "invalid-enum-value"`, without any warning from the compiler– I'd think it would complain since I'm not going through `LineCap.t`. Is there any way to "hide" the string type from calling code, and force it to use the submodule?

af: I believe my latest commit has addressed this– it's a bit wordy but I added a module interface for each of the enum submodules. Now the compiler gives you an error if you pass a random string instead of one of the enum values.

src/gl/canvas2dRe.re
```diff
@@ -0,0 +1,140 @@
...
+  let destinationOver : t;
+  let destinationIn : t;
+  let destinationOut : t;
+  let destinationAtop : t;
+  let lighter : t;
+  let copy : t;
+  let xor : t;
+};
+
+module Composite = {
```
glennsl: This doesn't use the CompositeType signature

src/gl/canvas2dRe.re
glennsl: I think `type t = private string`  would work too. But the proper way to do it I think would be to add an interface file to define the full interface of the entire module. Whatever you want to do, I'm OK to merge once you fix the `Composite` module

af: Thanks @glennsl, `type t = private string` is actually the first thing I tried (inside the sub-module definitions), but the compiler gives me `Unbound type constructor private`. AFAICT this is valid OCaml, but maybe Reason has different syntax for this?

The `Composite` module is also fixed now (a silly oversight, sorry!). If you're happy with it as-is, and I can't figure out the private type declaration (which admittedly would be a lot nicer than the verbose module types), it'd be nice to have this merged and then I can iterate in smaller PRs as needed.

glennsl: Apparently it's `type t = pri string;` in reason... You can use [reason-tools](https://github.com/reasonml/reason-tools) to easily convert between OCaml and Reason btw.

af:
@glennsl enum types are now private, as suggested. I got reason-tools set up now, this will be really useful! 

I don't know how much benefit there would be to put the full interface in a .rei file at this point, since almost everything is public and it'd be mostly duplication. But it's very much possible I'm missing something :) I'd vote to merge as-is, but happy to add the .rei file if you feel strongly about it. 

Thanks so much for all of your patience and tips with this PR btw. I've learned a lot!

glennsl:
Awesome! `.rei` files are very good for documentation too, but we can do without them for now. I'll go ahead and merge then. Thanks so much for doing all the work! :grinning:

PR #58, opened 2017-04-25
Change bs-platform from master to current release

saschatimme:


glennsl:
Sorry, I forgot about this. I've pinned `bs-platform` in the latest release, now published as `bs-webapi`. Master is still pinned for the moment, but is prone to track master again. I hope that'll do for you.

PR #62, opened 2017-05-26
Canvas gradient

bbqbaron:
First addition of any kind to these types, so I just wanted to make sure I wasn't missing anything obvious. Sorry for the overhead of a tiny PR, but hoping to follow this up with more Canvas bindings.

Next up I was going to tackle `createPattern`, which requires `HtmlImageElement`, which I assume introduces a new rabbithole. While I've got you: What causes the type `Dom.htmlElement` to exist? Still getting my feet under me with Ocaml module relationships. Is it somehow the line `module HtmlElement = HtmlElementRe`? That seems like it'd be self-referential, but I can't seem to find other ways that it could be declared.


bbqbaron:
Also, how do you all actually run your tests?

lib/js/tests/gl/canvas2d_test.js
```diff
@@ -30,9 +31,9 @@ ctx.lineJoin = Canvas2dRe.LineJoin[/* round */0];
 
 ctx.miterLimit = 10;
 
-ctx.strokeStyle = "red";
+ctx.strokeStyle = /* String */Block.__(0, ["red"]);
```
glennsl: This is why we have these "tests" and manually review them :)

src/gl/canvas2dRe.re
```diff
@@ -1,4 +1,6 @@
 type t;     /* Main type, representing the 2d canvas rendering context object */
+type gradientT;
```
glennsl: I'd prefer types to not have the `T` suffix. The `Gl` module is really the odd one out here as the only one that uses that convention.

src/gl/canvas2dRe.re
```diff
@@ -64,7 +66,9 @@ module LineJoin: LineJoinType = {
   let miter : t = "miter";
 };
 
-
+type style =
+  | String string
+  | Gradient gradientT;
```
glennsl: Could you try converting this into a GADT and using it as a phantom argument instead? See https://bloomberg.github.io/bucklescript/Manual.html#_phantom_arguments_and_ad_hoc_polymorphism. I think this should result in cleaner codegen. Another alternative is to use casting functions , e.g. `external styleFromString : string => style = "%identity"`

bbqbaron:
Ok! Changes applied. Looks like bobzhang beat me to the punch on opening an issue for needing to work around `[@@bs.set]` ([here](https://github.com/bloomberg/bucklescript/issues/1656)). Meantime, I've created type-annotated setter functions that just discard the constructor. The _test_ JS unfortunately no longer tells you anything useful, but the generated JS looks correct:

```
function setStrokeStyle(ctx, _, v) {
  ctx.strokeStyle = v;
  return /* () */0;
}

function setFillStyle(ctx, _, v) {
  ctx.fillStyle = v;
  return /* () */0;
}
```

bbqbaron:
Oh, it occurs to me that people could still find and call `fillStyle` and `strokeStyle` as-is, unsafely. Since presumably we'll be using a normal `[@@bs.set]` in future, is it worth creating an .rei to conceal them?

glennsl:
The GADT-fix looks great! I should check in the js artifacts too, but will wait until after this has been merged so you don't have to rebase :)

You can avoid exposing the externals by "shadowing" them, that is by naming the override functions the exact same as the externals. A neat trick for situations like this. Alternatively you could add an interface file and hand-pick what to expose, but that's a bit overkill at this point.

All the `bs.set` externals should be named `setX` too. And since they're actually read-write properties they should have getters. The convention is to have getters named `foo` and setters `setFoo`.

bbqbaron:
Great; will do tonight. Want me to sneak in the rest of the proposed canvas annotations, if it won't expand PR scope too much? Happy to follow up with a second PR on the heels of this one instead.

glennsl:
Great! I'm fine with whatever works for you :)

bbqbaron:
Okay; I added the rest of the Canvas bindings so we could handle any other issues all at once. I did run into one sticking point that I know of; will comment in-file.

src/gl/canvas2dRe.re
```diff
@@ -64,7 +68,14 @@ module LineJoin: LineJoinType = {
   let miter : t = "miter";
 };
 
+type image 'a =
+  | Number: image float
+  | ImageData: image ImageRe.t;
 
+type style _ =
```
bbqbaron: I may misunderstand something, but it seems like `style` needs a payload in its constructor, since we have a `fillStyle` getter that returns a valid `style` that we have to unwrap to get a value, right? This has an odd effect on `setFillStyle`. I figured it couldn't be `t => style 'a => 'a` anymore, since they'd have to call it with something like `ctx (String "whatever") "red"`, which seems weird. I converted it to switch on a single `style` value, but I can't seem to make the resulting `setXXX` external call _not_ use a `Block.__(0, ["red"])` in the JS. Will that outright break, or is the generated JS just weirder? It looks like it's still going to resolve to `"red"`, isn't it?

src/gl/canvas2dRe.re
```diff
@@ -88,20 +99,48 @@ external lineJoin : t => LineJoin.t => unit = "" [@@bs.set];
...
+};
+
+let setFillStyle (type a) (ctx: t) (s: style a) => switch (s) {
+  | (Gradient v) => setFillStyle ctx v
+  | (String v) => setFillStyle ctx v
+  | (Pattern v) => setFillStyle ctx v
+};
+
+let reifyStyle x => {
+  if (Js.typeof x == "string") {
```
bbqbaron: Oops, forgot to allow for Pattern after merging my other work. Fixing.

src/gl/canvas2dRe.re
glennsl: Varaints will be "boxed" in an array, so that unfortunately won't work properly. You'll have to use it as a phantom argument to be able to pass it directly as a value to the external. And the getter should then just return a tuple, like `reifyType`.

glennsl: Also, #1656 seems to be fixed, so you can try removing the overrides.

lib/js/tests/gl/canvas2d_test.js
```diff
@@ -30,9 +32,35 @@ ctx.lineJoin = Canvas2dRe.LineJoin[/* round */0];
...
+  case 2 : 
+      break;
+  
+}
+
+var match$1 = Canvas2dRe.strokeStyle(ctx);
+
+switch (match$1.tag | 0) {
+  case 0 : 
+      console.log(match$1[0]);
```
glennsl: You can see here how the GADT returned is an array with a tag property to switch on, and that the actual value is extracted by indexing into the array.

glennsl: Details here: https://bloomberg.github.io/bucklescript/Manual.html#_runtime_representation

lib/js/tests/gl/canvas2d_test.js
```diff
@@ -68,22 +96,96 @@ ctx.rect(0, 0, 10, 10);
...
 
 ctx.textBaseline = "top";
 
-ctx.fillText("hi", 1, 0);
+(function (eta) {
+      var param = /* None */0;
+      var param$1 = eta;
+      param$1.fillText("hi", 1, 0, param ? param[0] : undefined);
+      return /* () */0;
+    }(ctx));
```
glennsl: This is some weird code generation. cc @bobzhang (The offending declaration is `external fillText : string => x::float => y::float => maxWidth::float? => unit = "" [@@bs.send.pipe: t];`)

I think this is caused by not terminating an optional argument with a non-optional argument, confusing the compiler about whether or not to partially apply the function, but this codegen seems rather pointless. To fix it you can probably just add a unit argument, or have an "overload" `fillTextWithMaxWidth` to avoid the optional argument altogether.

src/gl/canvas2dRe.re
```diff
@@ -88,20 +99,48 @@ external lineJoin : t => LineJoin.t => unit = "" [@@bs.set];
...
+    String (Obj.magic x)
+  } else {
+    (Obj.magic (Gradient (Obj.magic x)))
```
glennsl: What's the first `Obj.magic` for?

You should also annotate the types whenever you use `Obj.magic` to keep the type inference form going off on its own.

lib/js/tests/gl/canvas2d_test.js
glennsl: Oh, I think it's an interaction with bs.send.pipe that kind of terminates the optional argument, resulting in this weirdness. lol

src/gl/canvas2dRe.re
bobzhang: Check out Js.Types

bbqbaron:
Okay, addressed almost everything that I know I need:
* `style` back to a content-free GADT
* ...returned in a tuple with its associated type from `reifyStyle`
* fixed `ImageData` constructor; i went with two explicit constructor functions rather than trying to pseudo-overload
* added some explanatory comments about pending fixes

Remaining:
* bs.set/ignore fix doesn't seem to be released yet, so I left it for now
* still need to gin up the raw instanceof, probably with just more directly raw code

src/gl/canvas2dRe.re
```diff
@@ -1,4 +1,10 @@
+/* don't refmt this file until refmt stops stripping bs.as. i don't know the issue number,
+  but have been told it's fixed-but-unreleased */
```
glennsl: BuckleScript uses refmt internally to convert from reason to ocaml, so it doesn't really matter whether or not you do it manually. The attributes will continue to be dropped until the fix has been released. But we should leave a comment here to remind ourselves to check that it's actually fixed once it does get released.

The issue is https://github.com/facebook/reason/issues/821

src/gl/canvas2dRe.re
```diff
@@ -88,20 +101,49 @@ external lineJoin : t => LineJoin.t => unit = "" [@@bs.set];
...
+external canvasGradient : 'a = "CanvasGradient" [@@bs.val];
+let instanceOf a cxr => [%bs.raw{|function(x,y) {return x instanceof y}|}] a cxr;
+
+let reifyStyle (type a) (x: 'a): (style a, a) =>
+  (if (Js.typeof x == "string") {
+    Obj.magic String
+  } else if (instanceOf x canvasGradient) {
+    Obj.magic Gradient
+  } else {
+    Obj.magic Pattern
```
glennsl: This assumes that these will be the only possible alternatives for the lifetime of this code. But since web standards tend to change I think it would be a good idea to run a check on `Pattern` too, and explicitly raise an `Invalid_argument`  exception if the unexpected happens. Just to be safe :)

src/gl/canvas2dRe.re
```diff
@@ -88,20 +101,49 @@ external lineJoin : t => LineJoin.t => unit = "" [@@bs.set];
...
+/* in re unused warnings
+  awaiting release of https://github.com/bloomberg/bucklescript/issues/1656
+  to just use [@@bs.set] directly with an ignored (style a) */
+let setStrokeStyle (type a) (ctx: t) (s: style a) (v: a) =>
+  setStrokeStyle ctx v;
+
+let setFillStyle (type a) (ctx: t) (s: style a) (v: a) =>
+  setFillStyle ctx v;
+
+external canvasGradient : 'a = "CanvasGradient" [@@bs.val];
```
glennsl: Could you mark these as internal?

bbqbaron:
w00t:
- [x] `bs.as` comment moved, corrected
- [x] `/* internal */` instance checks
- [x] handling `CanvasPattern` explicitly

glennsl:
Alright, I think we're good to merge then. Thank you!

PR #65, opened 2017-06-12
added getAttribLocation and drawArrays binding to webglre

tiansijie:


src/canvas/webGlRe.re
```diff
@@ -101,4 +101,8 @@ external enableVertexAttribArray : glT => int => unit = "enableVertexAttribArray
 /* void vertexAttribPointer(GLuint indx, GLint size, GLenum type,
    GLboolean normalized, GLsizei stride, GLintptr offset); */
 external vertexAttribPointer : glT => int => int => int => Js.boolean => int => int => unit =
-  "vertexAttribPointer" [@@bs.send];
\ No newline at end of file
+  "vertexAttribPointer" [@@bs.send];
+  /* GLint gl.getAttribLocation(program, name); */
+external getAttribLocation : glT => programT => string => int = "getAttribLocation" [@@bs.send];
+/* void gl.drawArrays(mode, first, count); */
+external drawArrays : glT => int => int => int => unit = "drawArrays" [@@bs.send];
```
glennsl: These bindings need to be completely revamped, so this is fine as it is in this context, but just for future reference:
* A function taking multiple arguments of the same type should have those arguments labeled to help distinguish them
* BuckleScript will infer the name function from the name of the external if given a blank string, e.g. `... = "" [@@bs.send]`

glennsl:
Thanks!

PR #67, opened 2017-06-13
Add geometry to bundle

bloodyowl:


glennsl:
Oh damn, that's a significant oversight. Thanks!

PR #71, opened 2017-07-07
Deprecate Storage, remove ReasonJs

chenglou:
@glennsl maybe a release before (or after) this?


glennsl:
Just rolled a release

PR #73, opened 2017-08-07
createText -> createTextNode

sidraval:
I believe `createText` should be `createTextNode`

 https://developer.mozilla.org/en-US/docs/Web/API/Document/createTextNode


glennsl:
Indeed it should. Thanks!

PR #74, opened 2017-08-14
Element focus

bloodyowl:


glennsl:
`focus` is on HTMLElement: https://github.com/reasonml-community/bs-webapi-incubator/blob/master/src/dom/html/htmlElementRe.re#L62

bloodyowl:
seems like `htmlElement` is confused with `htmlHtmlElement` https://github.com/reasonml-community/bs-webapi-incubator/blob/master/src/dom/html/htmlElementRe.re#L4-L5

glennsl:
Hmm, indeed. I guess I've always just used `Element.asHtmlElement`, so I've not noticed the error. I'll make a point release to fix this, thanks for pointing it out!

PR #75, opened 2017-08-14
Default view

bloodyowl:
Adds Document.defaultView


glennsl:
Hi! Sorry, I've missed these PRs somehow. But in any case, `defaultView` should be, and is, part of HTMLDocument, not Document: https://github.com/reasonml-community/bs-webapi-incubator/blob/master/src/dom/html/htmlDocumentRe.re#L9

PR #80, opened 2017-10-26
Adding file, image and URL definitions to window

arnarthor:
Adds bindings for File, Image and URL in the window global API

### DIFF
--- a/src/dom/html/WindowRe.re
+++ b/src/dom/html/WindowRe.re
@@ -95,6 +95,28 @@ module Impl (T: { type t; }) => {
 
 
   external setOnLoad : t_window => (unit => unit) => unit = "onload" [@@bs.set]; /* use addEventListener instead? */
+
+  module Image = {
+    type t;
+    external make : unit => t = "Image" [@@bs.new];
+    external src : t => string => unit = "" [@@bs.set];
+    external getSrc : t => string = "src" [@@bs.get];
+    external onload : t => (unit => unit) => unit = "" [@@bs.set];
+    external naturalHeight : t => string = "" [@@bs.get];
+    external naturalWidth : t => string = "" [@@bs.get];
+  };
+  
+  module File = {
+    type t;
+    external _type : t => string = "type" [@@bs.get];
+    external preview : t => string = "" [@@bs.get];
+  };
+  
+  module URL = {
+    external createObjectURL : File.t => string = "" [@@bs.val] [@@bs.scope "URL"];
+    external revokeObjectURL : string => unit = "" [@@bs.val] [@@bs.scope "URL"];
+  };
+
 };
 
 /* include WindowOrWorkerGlobalScope? not really "dom" though */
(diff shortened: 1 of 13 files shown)


src/dom/html/WindowRe.re
```diff
@@ -95,6 +95,28 @@ module Impl (T: { type t; }) => {
 
 
   external setOnLoad : t_window => (unit => unit) => unit = "onload" [@@bs.set]; /* use addEventListener instead? */
+
+  module Image = {
```
glennsl: This module really describes an `HTMLImageElement` (see https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/Image) and should therefore be in a module named accordingly.

Perhaps it should be aliased as `Image` somewhere to better conform to expectations, but I'm not really sure where it would make sense to place it.

src/dom/html/WindowRe.re
```diff
@@ -95,6 +95,28 @@ module Impl (T: { type t; }) => {
...
+    type t;
+    external make : unit => t = "Image" [@@bs.new];
+    external src : t => string => unit = "" [@@bs.set];
+    external getSrc : t => string = "src" [@@bs.get];
+    external onload : t => (unit => unit) => unit = "" [@@bs.set];
+    external naturalHeight : t => string = "" [@@bs.get];
+    external naturalWidth : t => string = "" [@@bs.get];
+  };
+  
+  module File = {
```
glennsl: This is part of the File API, not the DOM, and so should be in `src/file/FileRe.re` or something.

Also, while I understand that these implementations are incomplete, how would you even acquire a `File.t` instance?

src/dom/html/WindowRe.re
```diff
@@ -95,6 +95,28 @@ module Impl (T: { type t; }) => {
...
+    type t;
+    external _type : t => string = "type" [@@bs.get];
+    external preview : t => string = "" [@@bs.get];
+  };
+  
+  module URL = {
```
glennsl: Similarly, this API is not part of the DOM either. As joy would have it, it's part of both the `FIle API` and `URL` specs, and these function specifically are introduced in the File API, so perhaps put these in `src/file/` too.

arnarthor:
Yup, comments make perfect sense. I just copied this straight from our internal project where it was in window.re because these values are hooked to ex. `new window.Image(...)` 

I'll fix this 👍

src/dom/html/WindowRe.re
arnarthor: This is inferred in our project when using a file input where we send the type into a function that accepts a File.t

```ocaml
let file = List.hd (ReactDOMRe.domElementToObj (ReactEventRe.Form.target event))##files;
```

src/dom/html/WindowRe.re
glennsl: Did you forget to remove these from `Window`?

arnarthor: I sure did! Should be ready now 👍

glennsl:
Thanks!

PR #82, opened 2017-10-29
Use namespaces

arnarthor:
This enables namespaces, but there is a problem though, which is that bs-platform ships with a module called Dom so renaming DomRe.re to Dom.re is not possible because bs-platform does not have namespaces set right now. By renaming DomRe to Dom the compiler complains about circular dependency Dom -> AnimationEventRe -> Dom, which is untrue because AnimationEventRe is refering to the bs-platform Dom module , but the compiler doesn't seem to figure that out. I don't know if there's some way to pin this to a specific module (I guess not based on my understanding on the module system).

I don't know how to prevent this without enabling namespaces in bs-platform.

What's your take on this @glennsl? Enabling namespaces in bs-platform would of course be ideal, but I guess that's something that requires a bit more work than just turning on the knob.

/cc @bobzhang @chenglou should bs-platform enable namespaces in its repo?


glennsl:
Would it work to keep `DomRe.re` for internal use, but add a `Dom.re` containing just `include DomRe`?

src/dom/html/WindowRe.re
```diff
@@ -83,6 +83,7 @@ module Impl (T: { type t; }) => {
   external print : unit = "" [@@bs.send.pipe: t_window];
   external prompt : string => string = "" [@@bs.send.pipe: t_window];
   external promptWithDefault : string => string => string = "prompt" [@@bs.send.pipe: t_window];
+  external requestAnimationFrame : (float => unit) => unit = "" [@@bs.val];
```
glennsl: This doesn't really belong in the Dom API. It's only attached to `window` because that's the global scope for historic reasons. It was originally introduced in the animation-timing standard, but is now part of the animation frames section of the HTML5 standard. Maybe add an `Animation` module and dump it there instead.

arnarthor: Sounds good, I'll fix it 👍

arnarthor:
Having Dom.re with `include DomRe` unfortunately doesn't work. I tried that before submitting with no luck 😕

glennsl:
Hmm, that's unfortunate. I don't know how to do this without breaking backwards compatibility with something. [Namespace prelude](https://github.com/BuckleScript/bucklescript/issues/2215) would probably help, but of course that's not an option yet.

glennsl:
@arnarthor @bobzhang is this not basically the same issue we have here?
 https://github.com/BuckleScript/bucklescript/issues/2249. Does the resolution of that also fix this?

arnarthor:
I pretty sure this is the same issue yes 👍. I can try compiling with latest bucklescript master to verify tomorrow 👍

yawaramin:
This is now effectively done by #137

PR #85, opened 2017-11-19
Add HTMLFormElement & HTMLInputElement bindings

alexfedoseev:
This is almost ready, but still few things need to be done. Few questions before I can continue:

1. I use 2 types — `Dom.htmlFormElement` & `Dom.htmlInputElement` — that don't exist in BS's `Dom` namespace. How should this be handled?
1. Some externals are applicable only to elements w/ specific value of specific attribute (e.g. only type "checkbox" or "radio"). Should this be handled somehow?
1. JS's `NaN` is of `float` type in Reason, right?

P.S. Btw is there any plans to merge `Webapi`'s `Dom` and BS `Dom`? It's quite confusing that these are separate namespaces and it makes it harder to add required types (like `Dom.htmlFormElement` & `Dom.htmlInputElement`).


glennsl:
Awesome! To answer your questions:

1. The types can be put in `DomTypes` for now, until the next `bs-platform` release that contain the types.
2. It probably should, but I'm not sure how. Separate types for each would be the proper way, but not sure how possible or convenient that is.
3. Yes, `float` would be the correct type (`NaN` is an IEEE 754 floating point value)

The BS `Dom` module was actually (quite intentionally) factored out of `bs-webapi` to improve library interop without everything having to lug around this beast of a library. So while it's a bit inconvenient to develop `bs-webapi` it's very convenient for most everything else. It is confusing though, but I hope that can be fixed by movind BS `Dom` into the `Js` namespace.

alexfedoseev:
Oh crap, I completely forgot about this PR after I got back from vacations lol. I will finish it some time soon.

alexfedoseev:
Note to myself: don't forget to expose it before merge.

glennsl:
Seeing as this has been on ice for so long already, I don't mind waiting a bit longer :) Unless you really need it or something.

alexfedoseev:
Nope, no rush 👍

src/dom/html/HtmlInputElementRe.re
```diff
@@ -0,0 +1,123 @@
...
+
+  /* Properties that apply only to elements of type "file" */
+  [@bs.get] external accept : t_htmlInputElement => string = "";
+  [@bs.set] external setAccept : (t_htmlInputElement, string) => unit = "accept";
+  /* TODO: files: Returns/accepts a FileList object. */
+
+  /* Properties that apply only to text/number-containing or elements */
+  [@bs.get] external autocomplete : t_htmlInputElement => string = "";
+  [@bs.set] external setAutocomplete : (t_htmlInputElement, string) => unit = "autocomplete";
+  [@bs.get] external maxLength : t_htmlInputElement => int = "";
```
glennsl: The corresponding `minLength` seems to be missing

src/dom/html/HtmlInputElementRe.re
```diff
@@ -0,0 +1,123 @@
...
+  [@bs.set] external setMultiple : (t_htmlInputElement, bool) => unit = "multiple";
+  /* TODO: files: FileList array. Returns the list of selected files. */
+  [@bs.get] external labels : t_htmlInputElement => array(Dom.nodeList) = "";
+  [@bs.get] external step : t_htmlInputElement => string = "";
+  [@bs.set] external setStep : (t_htmlInputElement, string) => unit = "step";
+  [@bs.get] [@bs.return nullable] external valueAsDate : t_htmlInputElement => option(Js.Date.t) = "";
+  [@bs.set] external setValueAsDate : (t_htmlInputElement, Js.Date.t) => unit = "valueAsDate";
+  [@bs.get] external valueAsNumber : t_htmlInputElement => float = "";
+
+  [@bs.send.pipe: t_htmlInputElement] external focus : unit = "";
```
glennsl: `focus`, `blur` and `click` should get inherited from `HtmlElement`

src/dom/html/HtmlInputElementRe.re
```diff
@@ -0,0 +1,123 @@
...
+  [@bs.send.pipe: t_htmlInputElement] external blur : unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external select : unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external click : unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external setSelectionRange : (int, int) => unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external setSelectionRangeWithDirection : (int, int, string) => unit = "";
```
glennsl: This is missing an override of the name. The `string` argument here should ideally also be a variant.

src/dom/html/HtmlInputElementRe.re
```diff
@@ -0,0 +1,123 @@
...
+  [@bs.send.pipe: t_htmlInputElement] external setRangeTextWithinSelection : string => unit = "setRangeText";
+  [@bs.send.pipe: t_htmlInputElement] external setRangeTextWithinInterval : (int, int, string) => unit = "setRangeText";
+  [@bs.send.pipe: t_htmlInputElement] external setCustomValidity : string => unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external checkValidity : bool = "";
+  [@bs.send.pipe: t_htmlInputElement] external stepDown : unit = "";
```
glennsl: These can optionally take an int with the number of steps to take.

src/dom/html/HtmlInputElementRe.re
```diff
@@ -0,0 +1,123 @@
...
+  [@bs.send.pipe: t_htmlInputElement] external focus : unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external blur : unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external select : unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external click : unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external setSelectionRange : (int, int) => unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external setSelectionRangeWithDirection : (int, int, string) => unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external setRangeTextWithinSelection : string => unit = "setRangeText";
+  [@bs.send.pipe: t_htmlInputElement] external setRangeTextWithinInterval : (int, int, string) => unit = "setRangeText";
+  [@bs.send.pipe: t_htmlInputElement] external setCustomValidity : string => unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external checkValidity : bool = "";
```
glennsl: `reportValidity` seems to be missing

src/dom/html/HtmlInputElementRe.re
```diff
@@ -0,0 +1,123 @@
...
+  [@bs.get] external valueAsNumber : t_htmlInputElement => float = "";
+
+  [@bs.send.pipe: t_htmlInputElement] external focus : unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external blur : unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external select : unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external click : unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external setSelectionRange : (int, int) => unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external setSelectionRangeWithDirection : (int, int, string) => unit = "";
+  [@bs.send.pipe: t_htmlInputElement] external setRangeTextWithinSelection : string => unit = "setRangeText";
+  [@bs.send.pipe: t_htmlInputElement] external setRangeTextWithinInterval : (int, int, string) => unit = "setRangeText";
```
glennsl: There's also an extra optional `SelectionMode` argument here.

src/dom/html/HtmlInputElementRe.re
alexfedoseev: Fixed.

src/dom/html/HtmlInputElementRe.re
alexfedoseev: Fixed.

src/dom/html/HtmlInputElementRe.re
alexfedoseev: Fixed.

src/dom/html/HtmlInputElementRe.re
```diff
@@ -0,0 +1,123 @@
...
+  [@bs.set] external setType : (t_htmlInputElement, string) => unit = "type";
+  [@bs.get] external disabled : t_htmlInputElement => bool = "";
+  [@bs.set] external setDisabled : (t_htmlInputElement, bool) => unit = "disabled";
+  [@bs.get] external autofocus : t_htmlInputElement => bool = "";
+  [@bs.set] external setAutofocus : (t_htmlInputElement, bool) => unit = "autofocus";
+  [@bs.get] external required : t_htmlInputElement => bool = "";
+  [@bs.set] external setRequired : (t_htmlInputElement, bool) => unit = "required";
+  [@bs.get] external value : t_htmlInputElement => string = "";
+  [@bs.set] external setValue : (t_htmlInputElement, string) => unit = "value";
+  /* TODO: validity: ValidityState object. Returns the validity state that this element is in. */
```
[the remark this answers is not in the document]
alexfedoseev: AFAIU `ValidityState` is the whole different module which is not yet implemented? Or am I missing something?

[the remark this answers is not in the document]
alexfedoseev: Agreed, I didn’t look into it at all tbh. Will add.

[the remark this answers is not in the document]
alexfedoseev: Done.

alexfedoseev:
@glennsl Ping. It prolly safe to merge/release?

glennsl:
Indeed, we good now. Thanks! :smile:

PR #90, opened 2018-01-03
Move HtmlDocument.readyState to Document.readyState

abdullahibnnadjo:
Hi, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState), the readyState attribute belongs to the `Document` interface, rather than the `HtmlDocument` one, so just make it accessible from there.

However, I don't know how to run tests?

### DIFF
--- a/lib/js/src/dom/html/HtmlDocumentRe.js
+++ b/lib/js/src/dom/html/HtmlDocumentRe.js
@@ -22,9 +22,6 @@ function Impl() {
     self.dir = DomTypesRe.encodeDir(value);
     return /* () */0;
   };
-  var readyState = function (self) {
-    return DomTypesRe.decodeReadyState(self.readyState);
-  };
   var execCommand = function (command, show, value, self) {
     return +self.execCommand(command, Js_boolean.to_js_boolean(show), Js_null.from_opt(value));
   };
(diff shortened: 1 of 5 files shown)


glennsl:
I'm afraid MDN confuses this a bit. You can see on the bottom of the page you link to that it refers to the HTML spec. You can also see on the [Document](https://developer.mozilla.org/en-US/docs/Web/API/Document) page that it is listed in the "Extension for HTML document" section.

See also the [definition of the Document interface in the DOM spec](https://dom.spec.whatwg.org/#interface-document), and [the HTML spec](https://html.spec.whatwg.org/multipage/dom.html#document), which of course are the definitive sources for this.

I'll therefore close this, but thanks for making the effort to try to fix it anyway. More eyes are always better :)

abdullahibnnadjo:
Oh you're right! I needed to reread that a couple of times to get it.
If not being careful, the fact that it's listed on the `Document` interface is very misleading. It's even said:
> The Document.readyState property of a document […]

And on the HTML spec, I thought there were a `Document` interface with another separate `HTMLDocument` one, but hey, it's in fact the same interface name `Document`, but extended from the one in the DOM spec. The key part is:
> The WHATWG DOM standard defines a Document interface, which this specification extends significantly.

Which I completely missed, and didn't realize that there were different spec.
___
This is where OCaml strong type system (with the webapi 👍) is very useful to deal with the DOM, as it's very easy to make mistakes when working with it in normal JS (like this one…)

PR #91, opened 2018-01-03
Support bsb-native target js builds

lpalmes:
This will allow the lib to only compile when bsb is using js as the backend.
I had a problem when using this lib in a bsb-native project, props to @bsansouci for giving me this solution.

### DIFF
--- a/bsconfig.json
+++ b/bsconfig.json
@@ -5,6 +5,7 @@
 {
   "name": "bs-webapi",
   "bsc-flags": ["-bs-no-version-header", "-bs-super-errors"],
+  "allowed-build-kinds": "js",
   "sources": [
     {
       "dir": "src",


glennsl:
While this isn't a very intrusive change, I'm a bit uneasy about setting a precedent for solving problems like this in the wrong place.  I certainly wouldn't want every library to have to add this just in case they're used in native builds.

@bsansouci Wouldn't something like "bs-native-dependencies" and "bs-js-dependencies" keys make more sense?

@lpalmes If you really need this to get unblocked I'll merge it, but I'd really rather not, and it seems you should be able to get away with a temporary fork or something. Let me know if you can't though.

lpalmes:
I can work around this, with a a small fork of my own. I was publishing this pr to make it available for future uses in bsb-native. thanks for you concern

glennsl:
Thanks, I'll close this then. But feel free to keep discussing the issue here.

bsansouci:
@glennsl You have a point. I too don't want to have JS people have to change things specifically for native users. Though we have thought about it and this was the best solution given the tradeoffs.
We assumed most packages would build to all 3, so bob-native defaults to thinking it can build a package to all backends. We also don't allow the library user to tell bsb-native if a library can build to a target because it doesn't seem like the right order, the library should tell you where it can be built.
@lpalmes We'll add the ability to override the default from the library user though, as an escape hatch and maybe consider changing the default to only JS. 
The last thing is a breaking change so we'd have to consider it more.

glennsl:
Building to all targets seems like a good default I think, but library users will still need to be aware of which targets their dependencies support since they probably want to use it and have it work. They might also want to to build a dependency for only some of the targets it actually supports.

I think it's good for libraries to able to hint at what they support, so bsb can error out with a good explanation when incompatiblities are detected, but it shouldn't be required. What to build where is a consumer decision, not a library decision IMO. And it should be explicit so it doesn't cause confusion by magic.

PR #93, opened 2018-02-13
Fixing compile errors with bs-platform 2.2.0

hlysig:
Instead of using `Js.Null.to_opt`, we use `Js.Null.toOption`. Without this, compilation fails when using bs-platform 2.2.0.


glennsl:
Thanks! Unfortunately you've also heavily reformatted the modified files, so I'll apply this separately instead to reduce noise. But thanks for the heads-up!

hlysig:
Any time :)

PR #97, opened 2018-06-02
Add onerror to HtmlImageElementRe

wtfleming:
Allows writing code like this when loading an image:

```reason 
  let imageEl = HtmlImageElementRe.make();

  let myPromise =
    Js.Promise.make((~resolve, ~reject) => {
      HtmlImageElementRe.onload(imageEl, () => resolve(. imageEl));
      HtmlImageElementRe.onerror(imageEl, () =>
        reject(. Invalid_argument("could not load image"))
      );
    });

  HtmlImageElementRe.src(imageEl, "./tiles16.png");

```

### DIFF
--- a/src/dom/html/HtmlImageElementRe.re
+++ b/src/dom/html/HtmlImageElementRe.re
@@ -5,5 +5,6 @@ type t;
 [@bs.set] external src : (t, string) => unit = "";
 [@bs.get] external getSrc : t => string = "src";
 [@bs.set] external onload : (t, unit => unit) => unit = "";
+[@bs.set] external onerror : (t, unit => unit) => unit = "";
 [@bs.get] external naturalHeight : t => string = "";
 [@bs.get] external naturalWidth : t => string = "";


glennsl:
Hi. So first of all, you should be using `WebApi.Dom.HtmlImageElement` instead of `HtmlImageElementRe` as the latter is an internal module identifier that will be removed (or hidden, rather) when a decent build system happens to support it.

Secondly, it should pass the callback function a `Dom.event`, like `addEventListener` does.

Thirdly, this, and `onload`, should really be in a `GlobalEventHandlers` functor that's included with every element.

And (but?) finally, this should have inherited from `EventTarget`, which provides `addEventListener` and would have made this possible without adding new bindings. I also don't think `onerror` or `onload` provide any real value over `addEventListener`, but please correct me if I'm wrong. If I'm right, however, `onload` should instead be deprecated and removed. Then, optionally, `addLoadEventListener` and `addErrorEventListener` could be added to a new `GlobalEventHandlers` functor, but I'm not sure that adds much value either.

I'll close this for now, but please let me know if this doesn't work for you, if I'm wrong and this actually would provide value, or if you have any other questions. And let me know if you'd like to submit a PR for deprecating `onload` and adding `EventTarget` and the other inherited implementations to this. Otherwise I'll do it myself in the near future.

wtfleming:
Thanks so much for the info!

I'm a server side engineer by day, right now any Reason code I work on is pretty much just weekend projects for fun as I have time (I don't have a ton of familiarity with the DOM, and to be honest didn't even realize `addEventListener` existed).

The way I got to the example I wrote above was by trying to convert some of the code at https://developer.mozilla.org/en-US/docs/Games/Techniques/Tilemaps to Reason. The example I was following was directly setting onload and onerror instead of using `addEventListener`. A github search in the project for 'onload' led me to `HtmlImageElementRe` and since it worked I just made the assumption I was in the right place and onerror not existing was an oversight.

I'm not sure if I am comfortable enough with Reason and this library yet to work on a PR for deprecating onload and adding EventTarget and the other inherited implementations to this. I'll let you know if I get ambitious and give it a shot, but feel free to take care of it yourself.

thanks for all the work you've done on this library!

glennsl:
No worries, this isn't the easiest library to contribute to (or just use, for that matter). And there's definitely an oversight here, but I think it's a different one :) But in any case, thanks for the heads up!

I'll try to get around to this soon, but in the meantime you should be able just copy this binding into your own project as long as you replace `t` with the fully qualified type.

glennsl:
Just got around to this and have released a new version with these changes and a few more additions.

PR #98, opened 2018-06-21
Make selection nodes of nullable type

alexfedoseev:
https://www.w3.org/TR/selection-api/#dom-selection-anchornode


glennsl:
:+1:

PR #104, opened 2018-07-27
Make some properties of Element + Window floats

### LINKED ISSUE #102: type signatures for pixel count accessors
in the event of the user increasing / decreasing the level of zoom in their browser, `Dom.Window.scrollY`, `Dom.Element.offsetTop`, etc. will return return floats rather than integers. Current type signatures make it difficult to account for this in a sane manner.

Chrome 67.0.3396.99 / Firefox 60.0.1 - OS X 10.13.3

zploskey:
These should be floats based on the spec: https://drafts.csswg.org/cssom-view/

Fixes #102.


src/dom/html/WindowRe.re
```diff
@@ -44,17 +44,17 @@ module Impl = (T: {type t;}) => {
   [@bs.get] [@bs.return nullable] external opener : t_window => option(Dom.window) = "";
   [@bs.get] external outerWidth : t_window => int = "";
   [@bs.get] external outerHeight : t_window => int = "";
-  [@bs.get] external pageXOffset : t_window => int = ""; /* alias for screenX */
-  [@bs.get] external pageYOffset : t_window => int = ""; /* alias for screenY */
+  [@bs.get] external pageXOffset : t_window => float = ""; /* alias for scrollX */
+  [@bs.get] external pageYOffset : t_window => float = ""; /* alias for scrollY */
```
zploskey: These are actually aliases of `scrollX` and `scrollY`, so I corrected the comments.

glennsl: Oops. Thanks :)

glennsl:
Looks good to me. Thank you :)

Issue #101, opened 2018-07-19
offsetParent should return an option?


stuartkeith:
`offsetParent` can return null (a common situation is recursively walking up the DOM until `offsetParent` returns `null` - i.e. the `body` element has been reached).

However, the binding of `offsetParent` does not return an option - so this case is not detectable and can cause an exception.

The following fixed it for me:

`[@bs.get] [@bs.return nullable] external offsetParent : HtmlElementRe.t => option(Dom.element) = "";`

Thanks!

glennsl:
Indeed. Would you like to make a PR for it?

PR #105, opened 2018-08-01
Make offsetParent of nullable type

stuartkeith:
See #101


glennsl:
Thanks! I'll wait t release this until I get another PR though, but shouldn't be long

PR #109, opened 2018-09-09
Add URL and URLSearchParams

bryanthomaschen:
More fully implement [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) and [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URL)

Please feel free to update the module name mappings if necessary.

### DIFF
--- a/src/Webapi.re
+++ b/src/Webapi.re
@@ -12,3 +12,5 @@ module File = {
   include FileRe;
   module Url = UrlRe;
 };
+
+module Url = UrlRe;
--- a/src/file/URLSearchParamsRe.re
+++ b/src/file/URLSearchParamsRe.re
@@ -1,0 +1,15 @@
+[@bs.deriving abstract]
+type t;
+
+[@bs.new] external make: string => t = "URLSearchParams";
+[@bs.send] external append: (t, 'a, 'b) => unit = "append";
+[@bs.send] external delete: (t, 'a) => unit = "delete";
+[@bs.send] external entries: t => Js.Array.array_like('a) = "entries";
+[@bs.send] external get: (t, 'a) => option('b) = "get";
+[@bs.send] external getAll: (t, 'a) => Js.Array.array_like('a) = "getAll";
+[@bs.send] external has: (t, 'a) => bool = "has";
+[@bs.send] external keys: t => array('a) = "keys";
+[@bs.send] external set: (t, 'a, 'b) => unit = "set";
+[@bs.send] external sort: t => unit = "sort";
+[@bs.send] external toString: t => string = "toString";
+[@bs.send] external values: t => Js.Array.array_like('a) = "values";
--- a/src/file/UrlRe.re
+++ b/src/file/UrlRe.re
@@ -1,2 +1,19 @@
+[@bs.deriving abstract]
+type t = {
+  host: string,
+  hash: string,
+  hostname: string,
+  href: string,
+  origin: string,
+  pathname: string,
+  port: string,
+  protocol: string,
+  search: string,
+  searchParams: URLSearchParamsRe.t,
+};
+
+[@bs.new] external make: string => t = "URL";
+[@bs.new] external makeWithBase: (string, string) => t = "URL";
+
 [@bs.val] [@bs.scope "URL"] external createObjectURL : FileRe.t => string = "";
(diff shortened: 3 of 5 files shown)


src/file/UrlRe.re
```diff
@@ -1,2 +1,19 @@
...
+  host: string,
+  hash: string,
+  hostname: string,
+  href: string,
+  origin: string,
+  pathname: string,
+  port: string,
+  protocol: string,
+  search: string,
+  searchParams: URLSearchParamsRe.t,
```
glennsl: This should be read-only

src/file/UrlRe.re
```diff
@@ -1,2 +1,19 @@
...
+};
+
+[@bs.new] external make: string => t = "URL";
+[@bs.new] external makeWithBase: (string, string) => t = "URL";
```
glennsl: There's also a `toJSON` method that's missing: https://developer.mozilla.org/en-US/docs/Web/API/URL/toJSON

src/file/UrlRe.re
```diff
@@ -1,2 +1,19 @@
+[@bs.deriving abstract]
+type t = {
+  host: string,
+  hash: string,
+  hostname: string,
+  href: string,
+  origin: string,
```
glennsl: This should be read-only

src/file/URLSearchParamsRe.re
```diff
@@ -0,0 +1,15 @@
...
+type t;
+
+[@bs.new] external make: string => t = "URLSearchParams";
+[@bs.send] external append: (t, 'a, 'b) => unit = "append";
+[@bs.send] external delete: (t, 'a) => unit = "delete";
+[@bs.send] external entries: t => Js.Array.array_like('a) = "entries";
+[@bs.send] external get: (t, 'a) => option('b) = "get";
+[@bs.send] external getAll: (t, 'a) => Js.Array.array_like('a) = "getAll";
+[@bs.send] external has: (t, 'a) => bool = "has";
+[@bs.send] external keys: t => array('a) = "keys";
```
glennsl: This should be an `array_like` (or `iterator`, technically, but that concept does not exist yet in BuckleScript)

src/file/URLSearchParamsRe.re
```diff
@@ -0,0 +1,15 @@
+[@bs.deriving abstract]
+type t;
```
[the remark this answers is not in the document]
bryanthomaschen: It doesn't at the moment, since `URL.searchParams` returns a read-only record. I can remove it until there's a need to create such records.

src/Webapi.re
```diff
@@ -12,3 +12,5 @@ module File = {
   include FileRe;
   module Url = UrlRe;
 };
+
+module Url = UrlRe;
```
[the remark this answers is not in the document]
bryanthomaschen: It's just preference. =)
It `URL` doesn't seem to be part of `File`. But at the same time I didn't want to just move it in case it breaks stuff.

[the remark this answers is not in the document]
glennsl: I think it was put there because the parts that were implemented are defined in the File specification. But if this is added the old alias should be deprecated I think. I can do that myself though.

bryanthomaschen:
@glennsl Please let me know if there's anything else I can do. =)

src/file/UrlRe.re
```diff
@@ -1,2 +1,18 @@
...
+[@bs.new] external makeWithBase: (string, string) => t = "URL";
+[@bs.send.pipe : t] external host: string = "";
+[@bs.send.pipe : t] external hash: string = "";
+[@bs.send.pipe : t] external hostname: string = "";
+[@bs.send.pipe : t] external href: string = "";
+[@bs.send.pipe : t] external origin: string = "";
+[@bs.send.pipe : t] external pathname: string = "";
+[@bs.send.pipe : t] external port: string = "";
+[@bs.send.pipe : t] external protocol: string = "";
+[@bs.send.pipe : t] external search: string = "";
```
glennsl: Most of these should have setters. All except `origin` and `searchParams` I think. Would you mind adding adding those? Otherwise I can just do it when I merge.

bryanthomaschen: Done! I used `bs.set` and not `bs.send pipe` though. Just thought it'd be slightly clearer this way.

PR #111, opened 2018-09-17
Change bs.send.pipe to bs.get

### LINKED ISSUE #110: Bug with Url bindings
When doing:
```
let url = Webapi.Url.make("http://www.example.com");
Js.log(url->Webapi.Url.hostname);
```
The generated javascript is:
```
console.log(url.hostname());
```
This causes an error as the property is not a function.

Should change `[@bs.send.pipe : t]` to `[@bs.get]`

bryanthomaschen:
Fix #110


glennsl:
Thanks for the effort, but as pointed out in the issue this really isn't a bug. And even if this change was a good idea, it would have to be made across the entire library, which would break a lot of code for no reason.

glennsl:
Thanks!

PR #115, opened 2018-11-10
Add drag events

malthe:
This adds the following `Dom.dragEvent` types:
- `dragenter`
- `dragover`
- `dragleave`
- `drop`

### DIFF
--- a/src/dom/events/EventTargetRe.re
+++ b/src/dom/events/EventTargetRe.re
@@ -201,6 +201,64 @@ module Impl = (T: {type t;}) => {
   [@bs.send.pipe : T.t] external removeCompositionEndEventListener : ([@bs.as "compositionend"] _, Dom.compositionEvent => unit) => unit = "removeEventListener";
   [@bs.send.pipe : T.t] external removeCompositionEndEventListenerWithOptions : ([@bs.as "compositionend"] _, Dom.compositionEvent => unit, {. "capture": bool, "passive": bool}) => unit = "removeEventListener"; /* not widely supported */
   [@bs.send.pipe : T.t] external removeCompositionEndEventListenerUseCapture : ([@bs.as "compositionend"] _, Dom.compositionEvent => unit, [@bs.as {json|true|json}] _) => unit = "removeEventListener";
+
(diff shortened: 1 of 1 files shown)


malthe:
I must have used a wrong list. I'll add the remaining ones shortly.

In my own code, I did end up using this instead:
```ocaml
open Webapi.Dom

external __addEventListener : Element.t -> string -> ('a -> unit) = "addEventListener" [@@bs.send]

type dragEvent = [
  | `DragEnter [@bs.as "dragenter"]
  | `DragOver [@bs.as "dragover"]
  | `DragLeave [@bs.as "dragleave"]
  | `Drop [@bs.as "drop"]
]
[@@bs.deriving jsConverter]

type keyboardEvent = [
  | `KeyDown [@bs.as "keydown"]
  | `KeyUp [@bs.as "keyup"]
  | `KeyPress [@bs.as "keypress"]
]
[@@bs.deriving jsConverter]

type mouseEvent = [
  | `Click [@bs.as "click"]
  | `MouseDown [@bs.as "mousedown"]
  | `MouseUp [@bs.as "mouseup"]
  | `MouseMove [@bs.as "mousemove"]
]
[@@bs.deriving jsConverter]

let addEventListener element event listener =
  let open EventTargetRe in
  match event with
    `DragEnter
  | `DragOver
  | `DragLeave
  | `Drop as event -> __addEventListener element (dragEventToJs event) listener
  | `KeyDown
  | `KeyUp
  | `KeyPress as event -> __addEventListener element (keyboardEventToJs event) listener
  | `Click
  | `MouseDown
  | `MouseUp
  | `MouseMove as event -> __addEventListener element (mouseEventToJs event) listener
```
This is a more useful entry point in my case than what's offered in `EventTargetRe`.

glennsl:
Thanks, that would be great!

> This is a more useful entry point in my case than what's offered in EventTargetRe.

It's not type safe though.

malthe:
And true; that is a good point even though the way I have it set up, the add/remove calls are "managed" so that this compromise on type-safety (on the distinct events within each category) should work out okay in practice.

PR #117, opened 2018-12-02
This should be type 'Dom.attr', not 'Dom.element'

malthe:
See https://developer.mozilla.org/en-US/docs/Web/API/NamedNodeMap.

> The NamedNodeMap interface represents a collection of Attr objects.


glennsl:
Indeed. Thanks!

PR #118, opened 2018-12-08
Add 'popstate' event add/remove to window type

malthe:
I have included only the basic forms since to my knowledge, there is
no valid usecase for the extended ones in this particular case.


glennsl:
Looks good. Thanks!

PR #121, opened 2019-01-15
Add touch and animation events

malthe:
This adds events such as `touchstart` and `animationstart`.


glennsl:
Thanks! Looks good, except that you're missing `animationcancel`. See https://drafts.csswg.org/css-animations/#event-animationevent. Would you mind adding that?

Also, for future reference, it would help if you linked the sources you used for this, so I don't have to hunt around for them myself since they aren't always so easy to find.

malthe:
Done!

glennsl:
Thanks!

PR #123, opened 2019-02-05
Type correction for URLSearchParams

### LINKED ISSUE #122: URLSearchParams.get returns incorrect type
Currently the type of `get` in URLSearchParamsRe.re is:
`[@bs.send.pipe : t] external get: string => option(string) = "";`

As you can see in the documentation [here](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get)

The type I think should be:
`[@bs.send.pipe : t] external get: string => Js.Nullable.t(string) = "";`

Quite happy to make a PR if I am correct about this :)

SamWyld:
Fixes #122

### DIFF
--- a/src/file/URLSearchParamsRe.re
+++ b/src/file/URLSearchParamsRe.re
@@ -4,7 +4,7 @@ type t;
 [@bs.send.pipe : t] external append: (string, string) => unit = "";
 [@bs.send.pipe : t] external delete: string => unit = "";
 [@bs.send.pipe : t] external entries: Js.Array.array_like(string) = "";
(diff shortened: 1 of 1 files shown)


glennsl:
Thanks! It should be enough to add `bs.return` to it though, which preserves the better `option` type and has the additional benefit of not being a breaking change:

    [@bs.return nullable][@bs.send.pipe : t] external get: string => option(string) = "";

Could you try that instead?

SamWyld:
Yeah I agree. I have made the change 😄

glennsl:
Thanks!

PR #125, opened 2019-02-21
Add makeWithJson and makeWithObject to URLSearchParamsRe

feihong:
This corresponds with the use case where you want to create a query string using an object:
```reason
URLSearchParamsRe.makeWithObject({"foo": "hello", "bar": "world"})
  ->URLSearchParamsRe.toString;
```


feihong:
Yes, you're right. I've changed the functions to `makeWithDict` and `makeWithArray`.

glennsl:
Thanks!

PR #129, opened 2019-03-18
Added focusPreventScroll to HtmlElementRe

heklarun:
Added focusPreventScroll to HtmlElementRe

### DIFF
--- a/src/dom/html/HtmlElementRe.re
+++ b/src/dom/html/HtmlElementRe.re
@@ -71,6 +71,7 @@ module Impl = (T: {type t;}) => {
   [@bs.send.pipe : t_htmlElement] external blur : unit = "";
   [@bs.send.pipe : t_htmlElement] external click : unit = "";
   [@bs.send.pipe : t_htmlElement] external focus : unit = "";
+  [@bs.send.pipe : T.t] external focusWithOptions : {. "preventScroll": bool} => unit = "focus";
   [@bs.send.pipe : t_htmlElement] external forceSpellCheck : unit = ""; /* experimental */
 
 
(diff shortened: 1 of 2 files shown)


glennsl:
Ah, that's quite a few commits you got there :grinning: That's OK though, since the sum of all of them looks good I can just squash them when I merge so it'll end up as just one commit.

But for future reference, `git pull` will do a `fetch` followed by a `merge`, which will always just add commits. When you're out of sync with upstream, you will have to make destructive changes. If all you want is to make your local fork an exact mirror of upstream, you can do one of two things:

Either:

Delete both your fork on github and your local clone of that fork. Then go back here, push the fork button again to make a fresh fork on your account and clone that locally as usual. This will delete anything you've done in other branches too though.

Or, to update just the master branch:

If you haven't already, run:
```
git checkout master
git remote add upstream https://github.com/reasonml-community/bs-webapi-incubator
```

Then, to make sure your upstream reference is up to date:
```
git fetch upstream
```
This next step will overwrite everything you've done in the current branch with whatever is in the master branch of the upstream repository:
```
git reset --hard upstream/master
```
And then finally, to overwrite the branch on your github fork, you have to add `--force` to `git push`:
```
git push --force
```

Also, it's a good idea to make a separate branch for each PR. That way `master` won't get out of sync, it'll just be out of date, and you can simply do `git fetch upstream && git merge upstream/master` to add the latest commits to your master branch. No need to force push either, since you're not overwriting anything.

You'll get the hang of this soon enough, I'm sure. I struggled with exactly this for the first 10 or so PRs I made too. And it can add up to a lot more than this if you keep `pull`ing and doing everything on `master` :laughing:

Anyway, thanks for the PR! :smile:

Issue #130, opened 2019-03-20
Improve types of cloneNode and cloneNodeDeep


shtanton:
Currently `cloneNode` has the type signature:
`'a Dom.node_like -> Dom._baseClass Dom.node_like`

But with this signature, if I clone an element, it becomes a node, so I can't appendChild etc.

As I understand it we could either use a different return type for each module cloneNode appears in, or make use of the generic inheritance thing and change it to:
`'a Dom.node_like -> 'a Dom.node_like`

glennsl:
Indeed. `T.t` is an alias for the concrete type of each module, so returning that should be sufficient. Do you want to submit a PR?

PR #131, opened 2019-03-20
Refines cloneNode and cloneNodeDeep types

shtanton:
Both now return T.t, so clones of elements are also elements

Solves issue #130


glennsl:
Awesome, thanks!

PR #132, opened 2019-03-26
Remove 'pointless' window reference (which shadows the real reference)

malthe:
For ergonomics, it's nice to be able to do:
```ocaml
let open Webapi.Dom in
Window.innerWidth window
```
But this is only possible if `window` isn't defined as a property on `Window.t`. Of course, it _is_ such a property, but I don't think there's any valid use-case for accessing it.

### DIFF
--- a/src/Webapi/Webapi__Dom/Webapi__Dom__Window.re
+++ b/src/Webapi/Webapi__Dom/Webapi__Dom__Window.re
@@ -63,8 +63,6 @@ module Impl = (T: {type t;}) => {
   [@bs.get] external statusbar : t_window => statusbar = "";
   [@bs.get] external toolbar : t_window => toolbar = "";
   [@bs.get] external top : t_window => Dom.window = "";
-  [@bs.get] external window : t_window => t_window = ""; /* This is pointless I think, it's just here because window is the implicit global scope, and it's needed to be able to get a reference to it */
-
   [@bs.send.pipe : t_window] external alert : string => unit = "";
   [@bs.send.pipe : t_window] external blur : unit = "";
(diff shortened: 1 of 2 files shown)


glennsl:
Yeah, you're probably right. I'm considering moving `window`, `document` etc. into a `Current` module instead though, and perhaps provide a `Window` and `Document` module for shortcuts within that. That would allow you to write

```ml
let open WebApi.Dom in
Window.innerWidth Current.window
```

or just

```ml
let open WebApi.Dom in
Current.Window.innerWidth
```

The latter would require copying a lot of APIs almost verbatim, but it might be worth it since the current `window` or `document` is what you use more than 99% of the time. Perhaps it's possible to make this more manageable with som clever use of cppo.

What do you think?

In any case, this requires a period of deprecation first.

malthe:
Or `Webapi.Dom.Global` – to match `Js.Global`.

glennsl:
`Js.Global` is a different concept. It refers to the APIs that are defined in ECMAScript to belong to the "global" object, which might be `window` or something else, and should be available everywhere.

`Current` would refer to APIs that change depending on scope.

malthe:
Bump.

glennsl:
Hmm, I might be a little slow today, but I don't see how `Window.window` shadows anything in your example. `Webapi.Dom`window` might shadow a local variable, but the example doesn't show any local variable it might conflict with. So I'm unsure what the issue is here.

To be clear, I'm fine with removing `Window.window` since I can't see that it has any use, I'm just not sure it actually solves your issue.

malthe:
I don't really understand my own issue anymore – closing for now. Thanks.

PR #136, opened 2019-04-10
Enforce private modules

yawaramin:
Using BuckleScript's feature to only expose public modules.

### DIFF
--- a/bsconfig.json
+++ b/bsconfig.json
@@ -8,7 +8,8 @@
   "sources": [
     {
       "dir": "src",
-      "subdirs": true
+      "subdirs": true,
+      "public": ["Webapi"]
     },
     {
       "dir": "tests",


yawaramin:
I should mention that any modules _reachable_ from the toplevel module will also be public under both its aliased and original names, so it's not a foolproof solution. It's about the best we can do in OCaml right now, though.

glennsl:
Have you tested that this actually works from a consumer POV? Previously you had to list every single module that was to be included, even if it was only referenced by alias. I'm not aware that this has been changed.

Also, which modules would this actually hide? The only private module I can find is `DomInternalRe`, which is empty (I should really ask myself why it's still there...).

yawaramin:
Hi Glenn, it works. To test the happy path, I locally lifted out the `tests` subdirectory into a separate project, made it a consumer of `bs-webapi`, and ran the build (`bsb -make-world`). The build ran successfully. To test the sad path, I then created and set as public a dummy module (`Web.re`) in `bs-webapi-incubator`, then cleaned and ran the build in the test project again. This time the build failed with the expected errors, e.g.:

```
  We've found a bug for you!
  /Users/yawar/src/bs-webapi-incubator-test/src/dom/sets/domTokenList_test.re 1:6-15
  
  1 │ open Webapi.Dom;
  2 │ open DomTokenList;
  3 │ 
  
  The module or file Webapi can't be found.
  - If it's a third-party dependency:
    - Did you list it in bsconfig.json?
    - Did you run `bsb` instead of `bsb -make-world`
      (latter builds third-parties)?
  - Did you include the file's directory in bsconfig.json?
```

glennsl:
Alright, cool! Thanks for testing. As long as it now works as expected, it seems like a good thing.

I gather you don't need this right away then, so I'll merge but not push a release immediately.

yawaramin:
@glennsl thanks! That sounds good. I'll keep using `master`.

PR #137, opened 2019-04-14
Structure project in 'modular layout'

### LINKED ISSUE #81: Activate namespaces
Wouldn't it make senes to activate namespaces in this project? I see that the Bs_webapi module was deprecated in favour of Webapi but I still think it makes sense to activate namespaces here.

We're using this for some shimmed APIs in React Native but we needed to vendor this package to activate namespaces because both projects have a file called ImageRe.re

yawaramin:
Commit message follows. Sorry for the big-bang PR–adopting a new module structure is kind of an all-or-nothing proposition. As mentioned below, this change should not break end-users _as long as they have been using the aliases modules._ However, it does expose a couple of new modules that were not public before, so it should probably be at least in a minor release.

- Detailed discussion of the layout in
  https://dev.to/yawaramin/a-modular-ocaml-project-structure-1ikd
- Preserve the external-facing module hierarchy, e.g.
  `Webapi.Dom.Document`, `Webapi.Canvas.WebGl`, and all other modules
- Add a few modules to the public export aliases that are at least
  somewhat completed but still in the TODO list
- By using the BuckleScript `public` modules feature, and by not
  aliasing them, hide mixin modules like
  `Webapi_Dom_NonDocumentTypeChildNode`.

Fix #81


yawaramin:
It will be easier to review this PR starting from the bottom of the diff, the top 80% of it is JavaScript changes. Another good way to get a feel for it would be to just browse the directory structure of the branch and see the reorganized files.

src/Webapi/Webapi__Url.re
```diff
@@ -1,3 +1,22 @@
+module SearchParams = {
```
glennsl: Might be better  to keep this as `URLSearchParams` for consistency with the underlying object name. I'm open for convincing though.

src/Webapi.re
```diff
@@ -4,13 +4,13 @@ type rafId;
...
-module Dom = DomRe;
+module Base64 = Webapi__Base64;
+module Canvas = Webapi__Canvas;
+module Dom = Webapi__Dom;
 
 module File = {
-  include FileRe;
-  [@deprecated "Use `Webapi.Url` instead"] module Url = UrlRe;
+  include Webapi__File;
+  [@deprecated "Use `Webapi.Url` instead"] module Url = Webapi__Url;
```
glennsl: More of a note to self for later, perhaps, but I think this can be removed now.

lib/js/tests/Webapi/Webapi__Dom/Webapi__Dom__Window__test.js
```diff
@@ -16,7 +17,7 @@ window.crypto;
 
 window.document;
 
-window.frameElement;
+Caml_option.nullable_to_opt(window.frameElement);
```
glennsl: Hmm, this looks fishy. Are you using a really old version of BuckleScript here?

tests/Webapi/Webapi__Canvas/Webapi__Canvas__Canvas2d__test.re
```diff
@@ -1,6 +1,6 @@
 open Webapi.Canvas.Canvas2d;
 open Webapi.Canvas.CanvasElement;
-open Webapi.Dom;
+open! Webapi.Dom;
```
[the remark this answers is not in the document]
yawaramin: Good question, it's not actually needed if I delete the `module Image = Webapi__Dom__Image` binding from `Webapi__Canvas__Canvas2d.re`. I'll make that change.

src/Webapi/Webapi__Url.re
yawaramin: Ah, that's true. I'll change it back to that name.

lib/js/tests/Webapi/Webapi__Dom/Webapi__Dom__Window__test.js
yawaramin: The newest, as far as I know–6.0.0-dev.1. The handling of `[@bs.return nullable]` might have changed slightly.

yawaramin:
@glennsl regarding the fully-qualified directory names, I did it for predictability and discoverability. I wanted there to be a very simple and obvious pattern in all the file and directory names–e.g. the index module `Webapi__Dom.re` corresponds to the directory `Webapi__Dom` which contains all its submodules. I wanted to have that same pattern across all levels of the project.

lib/js/tests/Webapi/Webapi__Dom/Webapi__Dom__Window__test.js
glennsl: Might be better to use 5.x.x for this since the 6-series is pre-release, just in case there's any differences. But the generation does look right. `Caml_option` didn't exist until very recently either it seems.

glennsl:
Not sure how I feel about the directory naming scheme. There's a lot of redundancy and obfuscation with full paths: e.g. 
```
src/Webapi/Webapi__Dom/Webapi__Dom__Window.re
```
which could just be
```
src/dom/Webapi__Dom__Window.re
```

but I see your point, and it's really not so bad now that you've flattened the hierarchy. And since directory names aren't significant anyway and can easily be changed later, I'm okay leaving it as it is.

tests/Webapi/Webapi__Canvas/Webapi__Canvas__Canvas2d__test.re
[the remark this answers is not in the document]
glennsl: Maybe `Image.t` should be defined in `DomTypes` and aliased in Image, like most other `t`s are.

glennsl:
A few more questions:

* Which in-progress modules have you exported which weren't before?
* Could we signify that private modules are private through naming convention? Modules like `Webapi__Dom__Types` and `Webapi_Dom_NonDocumentTypeChildNode`. The latter shouldn't need to be namespace since it's properly private, but I'm uncertain about the Types module, since its types are aliased and exported, not just used in implementations.

tests/Webapi/Webapi__Canvas/Webapi__Canvas__Canvas2d__test.re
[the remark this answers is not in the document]
yawaramin: Sure, making that change as well.

lib/js/tests/Webapi/Webapi__Dom/Webapi__Dom__Window__test.js
yawaramin: OK, I've locally used v5.0.2 (the latest released version) and the `Caml_option` stuff is gone now.

glennsl: Oh, interesting, I didn't realize it was that new. Might even be a regression then.

yawaramin:
Regarding the directory naming scheme, let's give it a try for now, as you said. I have a feeling it will be easy to understand and unsurprising for people looking through the files.

The new exported modules are:

- `Webapi.Dom.NodeIterator`
- `Webapi.Dom.ProcessingInstruction`
- `Webapi.Dom.TreeWalker`

Regarding the private modules, I'm afraid I slightly misunderstood them. I just checked again in my dummy 'consumer' project and both `Webapi__Dom__Types` and `Webapi_Dom_NonDocumentTypeChildNode` are getting exported as public. I.e., their `.cmi`s show up in the compiled output, and they are usable from consumer code. So they're not private like we hoped, but fortunately because of the file names they shouldn't conflict with anything either.

glennsl:
Ah, I don't think those modules were intentionally left out. Probably just forgotten :)

yawaramin:
Gotcha, will do!

glennsl:
Released as 0.14. Thank you!

idkjs:
Nice notes, gentleman. Thank you.

PR #141, opened 2019-04-20
Add `window.getSelection`

pietro909:
As per [documentation on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection).

This is my first ReasonML code ever, I've started learning it yesterday evening: please take the time to review it carefully! It works _on my computer_ :-)

### DIFF
--- a/src/Webapi.re
+++ b/src/Webapi.re
@@ -1,8 +1,23 @@
 type rafId;
+type selection = {
+  _type: string,
+  anchorNode: Dom.element,
+  anchorOffset: int,
+  baseNode: Dom.element,
+  baseOffset: int,
+  extentNode: Dom.element,
+  extentOffset: int,
+  focusNode: Dom.element,
+  focusOffset: int,
+  isCollapsed: bool,
+  rangeCount: int,
+  toString: unit => string
+};
 
 [@bs.val] external requestAnimationFrame : (float => unit) => unit = "";
 [@bs.val] external requestCancellableAnimationFrame : (float => unit) => rafId = "requestAnimationFrame";
 [@bs.val] external cancelAnimationFrame : rafId => unit = "";
+[@bs.val] external getSelection : unit => selection = "";
 
 module Base64 = Webapi__Base64;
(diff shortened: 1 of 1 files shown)


pietro909:
Thanks! I completely missed the Window definition :-(
I have a question: why do we have `requestAnimationFrame` in this file then?  

About

> I have a hard time seeing how it would work on your machine either.

This is what happens if I comment out `toString` from the record definition:

```
$ bsb -make-world

[2/2] Building src/Webapi.mlast.d
[1/1] Building src/Webapi.cmj
ninja: no work to do.
[1/1] Building src/Content-HighlighterBs.cmj

  We've found a bug for you!
  /Users/pietrograndi/workspace/highlighter-bs/src/Content.re 7:42-49

  5 │
  6 │ Document.addEventListener("mouseup", _ => {
  7 │   let selection: string = getSelection().toString()
  8 │   if (String.length(selection) > 0) {
  9 │       chrome.extension.sendMessage({

  This expression has type Webapi.selection
  The field toString does not belong to type Webapi.selection
```

As I uncomment it, it works fine. I expected that _if it compiles, it works_, I haven't yet tried to run it (migrating some JS code from a personal project). Wouldn't expect a runtime error with a positive compilation -_-

glennsl:
> I expected that if it compiles, it works

When you write bindings, _you_ tell the compiler what the types are. The compiler can't know because external values are by definition foreign to it. And also constructs that exist at runtime only. You _could_ have an FFI system which ran checks on values at runtime to ensure they're the right type, but those checks could still only happen at runtime and the only cause of action would be to fail. So you'd get a runtime error anyway, just sooner.

Bottom line is, you have to be very diligent when writing bindings and make sure you get it right. Otherwise all bets are off. The type system only work on the assumption that bindings are correctly typed.

> why do we have requestAnimationFrame in this file then?

`requestAnimationFrame` isn't specific to any window instance. It's attached to the 'window' object only because that is defined as the global object, and it has to belong to something. There's lots of weirdness in Web APIs, unfortunately.

PR #142, opened 2019-04-28
Add Dom.Element height and width setters

drivasperez:
I spent a while trying to find these, and it doesn't seem like they're in the API?

New function bindings for Dom.Element: setHeight and setWidth, each taking an element and an integer, returning unit.


drivasperez:
Oh yes, I was using these for Canvas and had a brainfart there, I think. Would these fit in Dom.Canvas as `[@bs.set] external setHeight : (Webapi__Dom__Element.t, int) => unit = "height";` etc?

glennsl:
`Dom.Canvas.CanvasElement` would be the place I guess. Those are a bit loosely typed, but as long as they're there you might as well add a few more. Would you add getters too?

drivasperez:
Sure. I made a new PR at #143. Hopefully there's nothing embarrassingly wrong in it, but you never know.

PR #143, opened 2019-04-28
Add height and width getters and setters for Canvas elements

drivasperez:


src/Webapi/Webapi__Canvas.re
```diff
@@ -4,4 +4,8 @@ module WebGl = Webapi__Canvas__WebGl;
 module CanvasElement = {
   [@bs.send] external getContext2d : (Dom.element, [@bs.as "2d"] _) => Canvas2d.t = "getContext";
   [@bs.send] external getContextWebGl : (Dom.element, [@bs.as "webgl"] _) => WebGl.glT = "getContext";
+  [@bs.get] external height : Dom.element => int = "height";
```
glennsl: By convention these are just named `height` and `width`, without a `get` prefix.

glennsl:
Crap. Just the test itself that is fragile though, I'll fix locally and release shortly. Thanks!

src/Webapi/Webapi__Canvas.re
drivasperez: Ah, got it.

PR #145, opened 2019-05-06
Update DomRect

alexfedoseev:
`int` is changed to `float` b/c it's `unrestricted double` in specs:
https://www.w3.org/TR/geometry-1/#DOMRect
https://drafts.fxtf.org/geometry/#DOMRect


glennsl:
Dang, you right. Thanks!

PR #146, opened 2019-05-11
Add forEach to module URLSearchParams.

parkerziegler:
This PR adds the `forEach` method to the `URLSearchParams` module, as specified here: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/forEach. Users can now iterate over url search params as key-value pairs like so:

```reason
let params = URLSearchParams.make("key1=value1&key2=value2");
URLSearchParams.forEach(
   (value, key) => {
      print_endline(value);
      print_endline(key);
   },
   params
);

/*
value1
key1
value2
key2
*/
```

`refmt` also appears to have reformatted some of the lines in this module, but the addition of `forEach` is the only meaningful change here.

### DIFF
--- a/tests/Webapi/Webapi__Url__test.re
+++ b/tests/Webapi/Webapi__Url__test.re
@@ -1,0 +1,10 @@
+open Webapi.Url;
+
+let params = URLSearchParams.make("key1=value1&key2=value2");
+URLSearchParams.forEach(
+   (value, key) => {
+      print_endline(value);
+      print_endline(key);
+   },
+   params
+);
--- a/tests/Webapi/Webapi__Url__test.re
+++ b/tests/Webapi/Webapi__Url__test.re
@@ -1,10 +1,4 @@
 open Webapi.Url;
 
 let params = URLSearchParams.make("key1=value1&key2=value2");
-URLSearchParams.forEach(
-   (value, key) => {
-      print_endline(value);
-      print_endline(key);
-   },
-   params
-);
+URLSearchParams.forEach(Js.log2, params);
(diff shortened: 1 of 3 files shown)


parkerziegler:
Thanks @glennsl. Definitely agree with your changes, I'll make them later today when I get a moment and push up. I did run the code above and it does execute properly, but that callback will also get automatically curried, which we don't wan. `[@bs.uncurry]` seems perfect for this. Sorry for the commit noise, will also clean that up 😅 Didn't see any tests in place for this `module`, but happy to start some if that'd be helpful!

parkerziegler:
@glennsl changes should be made and commit noise cleaned up, let me know if you'd like any further changes!

tests/Webapi/Webapi__Url__test.re
```diff
@@ -0,0 +1,10 @@
+open Webapi.Url;
+
+let params = URLSearchParams.make("key1=value1&key2=value2");
+URLSearchParams.forEach(
+   (value, key) => {
+      print_endline(value);
```
yawaramin: You can use `Js.log2(value, key)` here to more closely follow the example in https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/forEach#Examples

yawaramin: I think this will work, actually:

```reason
URLSearchParams.forEach(Js.log2, params);
```

parkerziegler: Yeah I originally had `Js.log2` in there but went for `print_endline` to be more "Reason-y". But this is BuckleScript land so `Js.log2` makes more sense. And yep, that'll work perfectly. Pushing up now.

parkerziegler:
@glennsl anything else you'd like to see here? Thinking I'll add tests for the rest of this module in a follow up PR if that sounds useful.

glennsl:
Looks great. Thanks!

A few more tests would also be great. Individual tests for simple getters and setters aren't all that useful in themselves, but anything more complicated can have subtle issues, especially between compiler version.

PR #147, opened 2019-05-15
add scrollBy

r281GQ:
Hey guys!

First awesome lib! Thanks! I just wanted to create a component where I missed scrollBy. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollBy), what do you think about this?


glennsl:
Perfectly executed PR :smile: Thanks! Seems odd that this didn't already exist, but I guess it was just forgotten somehow.

PR #149, opened 2019-05-21
Add size method to Webapi.File module

leomayleomay:


src/Webapi/Webapi__File.re
```diff
@@ -4,4 +4,6 @@ type t;
 
 [@bs.get] external name : t => string = "name";
 
-[@bs.get] external preview : t => string = "";
\ No newline at end of file
+[@bs.get] external size : t => int = "size";
```
glennsl: Thanks! I think this needs to be a float though. `int` is considered 32-bit by BuckleScript while `float` is considered 64-bit. `int` will, if my math is correct, overflow if `size` is > 2GB. Which is unlikely, but certainly not impossible. See https://bucklescript.github.io/docs/en/difference-from-native-ocaml

leomayleomay: @glennsl good point, done the changes :)

PR #154, opened 2019-08-16
Added asNode

### LINKED ISSUE #150: Subtyping doesn't unify
I can't create a collection of various nodes like this:

```ocaml
let x =
  let document = Webapi.Dom.window |> Webapi.Dom.Window.document
  in
  [ Webapi.Dom.Document.createElement "foo" document
  ; Webapi.Dom.Document.createTextNode "bar" document ]
```

The subtyping fails, it doesn't compile

Risto-Stevcev:
Closes #150

### DIFF
--- a/src/Webapi/Webapi__Dom/Webapi__Dom__Node.re
+++ b/src/Webapi/Webapi__Dom/Webapi__Dom__Node.re
@@ -1,4 +1,6 @@
 module Impl = (T: {type t;}) => {
+  external asNode : T.t => Dom.node = "%identity";
+
   /* baseURI */
   [@bs.get] external childNodes : T.t => Dom.nodeList = "";
   [@bs.get] [@bs.return nullable] external firstChild : T.t => option(Dom.node) = "";


src/Webapi/Webapi__Dom/Webapi__Dom__Node.re
```diff
@@ -1,4 +1,6 @@
 module Impl = (T: {type t;}) => {
+  external asNode : Dom.node_like(_) => Dom.node = "%identity";
```
glennsl: The argument here should be `T.t` rather than `Dom.node_like(_)`. Otherwise you would be able to use `Document.asNode` on an `element`, for example, which might get confusing.

If you're thinking of having a generic catch-all `Node.asNode` function that'll work on any `node_like` it should be outside the `Impl` functor so it's not inherited. But I can't really see a use case for that, and would prefer strongly typed functions since it's less confusing and documents what exactly you're dealing with.

Risto-Stevcev: Well the use-case is: I want to have a collection of things that implement the `Node` interface, I think this is definitely useful and reasonable even if you disagree.. I ran into this issue because I needed it. `Dom.node_like` is for anything that implements the `Node` interface, so I don't to see how this coercion is unsafe or not strongly typed. :man_shrugging: 
I developed a subtyping system that's capable of being flexible enough to accommodate this without needing any coercion while still being strongly typed in `bs-declaredom`, but it would be a large undertaking to convert this library to use that style.

glennsl: "Strongly typed" was perhaps a poor choice of words. I don't mean it's unsafe, but that it might be confusing that it automatically coerces to a subtype, when nothing else does. I.e. that `Document.asNode(someElement)` compiles isn't unsafe, but it sure looks wrong.

So there are ways to fix this as I see it:

1. Keep `node_like` but move `asNode` out of the `Impl` functor. Then it wouldn't be inherited, `Document.asNode` wouldn't exist, and `Node.asNode` would accept any kind of node.

2. Replace `Dom.node_like(_)` with `T.t`, which is a placeholder for the type associated with the module. `Document.asNode(someElement)` would then not compile, neither would `Node.asNode(someElement)`, but `Element.asNode(someElement)` would.

Alternative 2 is consistent with the rest of the library, while alternative 1 is not. But they're not strictly equivalent since alternative 1 will accept _any_ subtype while alternative 2 needs explicit support for each subtype. So what I tried to say is that unless there's a use case for alternative 1, which I can't really see, we should go with alternative 2.

Risto-Stevcev: Yeah, I was originally thinking of doing it like #2, but would have to add it for all of the `node_likes`, like `Document.asNode(someDocument)`, `Element.asNode(someElement)`, `Text.asNode(someText)`, etc. What do you think?

glennsl: You don't have to add it anywhere else, it's inherited automatically by any module that includes `Node.Impl`. That's what the functor is for :)

Risto-Stevcev: Ah gotcha, updated it

PR #157, opened 2019-08-18
Add nodeList.forEach binding

### LINKED ISSUE #155: Nodelist forEach
Hello,

I'm new to ReasonML but I made search 
and it seams it lack forEach loop on Nodelist.

Currently, we have to use Array.forEach :
```ocaml
let items = Document.querySelectorAll(".item")(document);

Js.Array.forEach(item => Js.log(item), NodeList.toArray(items));
```

that output :
```javascript
var items = document.querySelectorAll(".item");

Array.prototype.slice.call(items).forEach((function (item) {
        console.log(item);
        return /* () */0;
      }));
```

With NodeList.forEach :
```ocaml
let items = Document.querySelectorAll(".item")(document);

Js.NodeList.forEach(item => Js.log(item), items);
```

that should output:
```javascript
var items = document.querySelectorAll(".item");

items.forEach((function (item) {
        console.log(item);
        return /* () */0;
      }));
```

dmartinjs:
Fix #155

### DIFF
--- a/src/Webapi/Webapi__Dom/Webapi__Dom__NodeList.re
+++ b/src/Webapi/Webapi__Dom/Webapi__Dom__NodeList.re
@@ -2,6 +2,8 @@ type t = Dom.nodeList;
 
 [@bs.val] external toArray : t => array(Dom.node) = "Array.prototype.slice.call";
 
+[@bs.send.pipe : t] external forEach : ((Dom.node, int) => unit) => unit = "";
+
 [@bs.get] external length : t => int = "";
 
 [@bs.send.pipe : t] [@bs.return nullable] external item : int => option(Dom.node) = "";
--- a/src/Webapi/Webapi__Dom/Webapi__Dom__NodeList.re
+++ b/src/Webapi/Webapi__Dom/Webapi__Dom__NodeList.re
@@ -2,7 +2,7 @@ type t = Dom.nodeList;
 
 [@bs.val] external toArray : t => array(Dom.node) = "Array.prototype.slice.call";
 
-[@bs.send.pipe : t] external forEach : ((Dom.element, int) => unit) => unit = "";
+[@bs.send.pipe : t] external forEach : ((Dom.node, int) => unit) => unit = "";
 
 [@bs.get] external length : t => int = "";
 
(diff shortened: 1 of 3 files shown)


src/Webapi/Webapi__Dom/Webapi__Dom__NodeList.re
```diff
@@ -2,7 +2,7 @@ type t = Dom.nodeList;
 
 [@bs.val] external toArray : t => array(Dom.node) = "Array.prototype.slice.call";
 
-[@bs.send.pipe : t] external forEach : ((Dom.node, int) => unit) => unit = "";
+[@bs.send.pipe : t] external forEach : ((Dom.element, int) => unit) => unit = "";
```
glennsl: Hmm, why did you do this? It's a NodeList after all, and while some functions like `querySelectorAll` will return only elements, `Node.childNodes` will not.

dmartinjs: [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach) only talk about elements.

And TypeScript have
```
NodeListOf<Element>.forEach(callbackfn: (value: Element, key: number, parent: NodeListOf<Element>) => void, thisArg?: any): void
```

Have you a better source for the type of this value ?

dmartinjs:
Done

src/Webapi/Webapi__Dom/Webapi__Dom__NodeList.re
glennsl: As I said, `Node.childNodes` returns a `NodeList` of all kinds of nodes. MDN is a wiki, you shouldn't trust it 100%. Look at the standards documents instead, like https://dom.spec.whatwg.org/#nodelist.

TypeScript seems to have parametetized it. What you're looking at is the type signature specifically for a `NodeList` of `Element`s. You could probably also have something like a `NodeListOf<Node>`, in which case the type of `forEach` would be

```
NodeListOf<Node>.forEach(callbackfn: (value: Node, key: number, parent: NodeListOf<Node>) => void, thisArg?: any): void
```

I'm sure we could do the same, but that's a bigger change. Give it a shot if you want.

PR #160, opened 2019-09-22
Add support for keypress events

msvbg:
Per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent), this is the last remaining keyboard event type.


msvbg:
Closing since it is deprecated https://developer.mozilla.org/en-US/docs/Web/API/Document/keypress_event.

TheSpyder:
I think it should be included. Webapi includes experimental or very new features not supported by older browsers, why not deprecated features as well for working with older browsers?

glennsl:
The difference is that deprecated features always have an alternative (unless there's a good reason for there not be an alternative, like a security problem), while experimental features provide new functionality that probably don't have a reliable alternate implementation.

I'm not entirely opposed to including deprecated features, but it needs to be further justified. Similar to how the experimental features included should be so far along in the standards process that the lack of stability is more of a formality.

TheSpyder:
In my case the use of a deprecated feature is because I'm using an experimental one on newer browsers. I'm working with the `beforeinput` event, one of the recommended alternatives to `keypress`, but it isn't even available on FireFox much less the Microsoft browsers. `keypress` is the only viable alternative; `keydown` is too restricted.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/beforeinput_event

For the moment I have both `beforeinput` and `keypress` as custom bindings on top of webapi, we had planned to make a PR for `beforeinput` at some point but if it's too experimental for webapi and we have to maintain it ourselves we might as well maintain `keypress` as well :)

glennsl:
In that case I think we should include both, but marked as deprecated and experimental respectively.

I find it really odd that functionality is deprecated before it's replacement is in place, though. That runs counter to what deprecation means, at least as I understand it.

@msvbg Do you want to reopen this?

TheSpyder:
The whole `beforeinput` saga is mind-blowing, to be honest 😅 but when working in browsers we make do with what we have.

I would appreciate adding both, thanks 👍 I'll make a PR for beforeinput soon.

msvbg:
I am fine either way! I am not using "keypress" however.

PR #161, opened 2019-09-23
selectionchange event

spocke:
Here is a implementation of the selectionchange event.

Specification:
https://w3c.github.io/selection-api/#selectionchange-event


spocke:
I added a new GlobalEventHandlers I hope it's what you where thinking. However the spec and the real world doesn't seem to align. The selection change event is only firing on document not on element, htmlelement or window. https://jsfiddle.net/nc290odu/ but according to the spec here it should be available on https://w3c.github.io/selection-api/#dom-globaleventhandlers-onselectionchange

So not sure what we do here do we follow the spec or or do we follow the real world. The first PR was more real world the new PR is more spec world. However they don't mention element so maybe the selectionevent shouldn't be available there.

PR #162, opened 2019-09-23
removed global conflict of None among other things

spocke:
When using `open Webapi.Dom` you get a conflict with the Option value constructor `None` and the Webapi `eventPhase None`. I think all othe Webapi__Dom__Types should be in either separate modules or in it's own module to avoid this open conflict. This PR is a breaking change that wraps the Types in it's own module, I think it would make the Webapi easier to use.


spocke:
Closing this awaiting feedback on how to proceed from main developer on original ticket.

PR #163, opened 2019-09-24
Less radical fix for Option None conflict

spocke:
This is a slightly less radical approach of fixing the Option None conflict by just wrapping the eventphase in a module.


spocke:
Closing this awaiting feedback on how to proceed from main developer on original ticket.

PR #165, opened 2019-11-14
add bindings to DomStringMap

tsnobip:
This is a small PR to add bindings to DomStringMap.
I know the rest of the API is t-last but I didn't find a way to use `bs.get_index` in a t-last way so these bindings are t-first, is this an issue? 
I wrote tests similar to the others, but I'm not sure how it's supposed to be run.

### DIFF
--- a/src/Webapi/Webapi__Dom/Webapi__Dom__DomStringMap.re
+++ b/src/Webapi/Webapi__Dom/Webapi__Dom__DomStringMap.re
@@ -1,0 +1,12 @@
+type t = Dom.domStringMap;
+
+type key = string;
+
+[@bs.get_index] [@bs.return nullable]
+external get: (t, key) => option(string) = "";
+let get = (key, map) => get(map, key);
+[@bs.set_index] external set: (t, key, string) => unit = "";
+let set = (key, value, map) => set(map, key, value);
+let unsafeDeleteKey: (key, t) => unit =  
+  [%raw (key, map) => "delete map[key];"];
+  
--- a/lib/js/src/Webapi/Webapi__Dom.js
+++ b/lib/js/src/Webapi/Webapi__Dom.js
@@ -1,6 +1,51 @@
 'use strict';
 
+var Webapi__Dom__Attr = require("./Webapi__Dom/Webapi__Dom__Attr.js");
+var Webapi__Dom__Node = require("./Webapi__Dom/Webapi__Dom__Node.js");
+var Webapi__Dom__Text = require("./Webapi__Dom/Webapi__Dom__Text.js");
 var Webapi__Dom__Types = require("./Webapi__Dom/Webapi__Dom__Types.js");
+var Webapi__Dom__Window = require("./Webapi__Dom/Webapi__Dom__Window.js");
+var Webapi__Dom__Comment = require("./Webapi__Dom/Webapi__Dom__Comment.js");
+var Webapi__Dom__Element = require("./Webapi__Dom/Webapi__Dom__Element.js");
+var Webapi__Dom__UiEvent = require("./Webapi__Dom/Webapi__Dom__UiEvent.js");
+var Webapi__Dom__Document = require("./Webapi__Dom/Webapi__Dom__Document.js");
+var Webapi__Dom__DragEvent = require("./Webapi__Dom/Webapi__Dom__DragEvent.js");
+var Webapi__Dom__TimeEvent = require("./Webapi__Dom/Webapi__Dom__TimeEvent.js");
+var Webapi__Dom__CloseEvent = require("./Webapi__Dom/Webapi__Dom__CloseEvent.js");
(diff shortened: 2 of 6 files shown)


src/Webapi/Webapi__Dom/Webapi__Dom_DomStringMap.re
```diff
@@ -0,0 +1,12 @@
...
+type key = string;
+
+[@bs.get_index] [@bs.return nullable]
+external get: (t, key) => option(string) = "";
+[@bs.set_index] external set: (t, key, string) => unit = "";
+let unsafeDeleteKey: (. t, key) => bool = [%raw
+  (map, key) => {|
+            return delete map[key]
+        |}
+];
```
glennsl: I don't think this really needs to return the `bool` value since it'll always be `true` in strict mode. Could you also try to format this on one line if it'll fit, or alternatively two with the entire `[%raw ...]` block on the second line?

I also think it should use t-last, despite being inconsistent with `get_index` and `set_index` it'll be consistent with every other function in `bs-webapi`. You could also obtain consistency internally by providing overrides for `get` and `set` that flip the argument order.

glennsl:
Thanks for the PR! See the comment above.

src/Webapi/Webapi__Dom/Webapi__Dom_DomStringMap.re
tsnobip: Thanks I didn't know `delete` behavior in strict mode, I guess I should learn JS more ^^
I wrapped `get` and `set` too, the inconsistency was too weird.
The `[% raw ...]` formatting was due to refmt that imo adds too many new lines indeed.

tsnobip:
Thanks @glennsl for reviewing the PR, I amended it according to your remarks, let me know if it's alright or not now.

tsnobip:
@glennsl we could also have a 0-cost t-last (and t-first) solution by using labels, plus I think it'd make the `set` function safer by not mixing the key and value that are of the same type. I think labels are not really used in bs-webapi either but the style change would be less unsettling than going from t-last to t-first. Let me know what you think of it.

src/Webapi/Webapi__Dom/Webapi__Dom__DomStringMap.re
```diff
@@ -0,0 +1,12 @@
...
+type key = string;
+
+[@bs.get_index] [@bs.return nullable]
+external get: (t, key) => option(string) = "";
+let getPipe: (key, t) => option(string) = (key, map) => get(map, key); 
+[@bs.set_index] external set: (t, key, string) => unit = "";
+let setPipe: (key, string, t) => unit = 
+  (key, value, map) => set(map, key, value);
+let unsafeDeleteKey: (. t, key) => unit =  
+  [%raw (map, key) => "delete map[key];"];
```
glennsl: Argument order needs to be swapped here. It would also be nice if this didn't use uncurrying. I think BuckleScript supports that for this construct.

src/Webapi/Webapi__Dom/Webapi__Dom__DomStringMap.re
```diff
@@ -0,0 +1,12 @@
+type t = Dom.domStringMap;
+
+type key = string;
+
+[@bs.get_index] [@bs.return nullable]
+external get: (t, key) => option(string) = "";
+let getPipe: (key, t) => option(string) = (key, map) => get(map, key);
```
glennsl: You can just name this `get`. Then it will shadow the `external`, but allows you to still use it inside the function definition.

tsnobip:
I'm really sorry @glennsl, it didn't pay attention and my latest changes didn't get pushed, like the shadowing and the argument order.
I'm going to use these bindings in a rich text editor library (prosemirror) where they can be called pretty often, but even then the extra cost is probably insignificant. I saw labelled arguments in a few places though, like in Canvas2d, DomRect or DomImage. 
But it's your call, I'm gonna push the version that really answers to your remarks tomorrow when I have access to my laptop.

glennsl:
`Canvas2d` is for historical reasons pretty inconsistent with the rest of `bs-webapi`, but even there I think labels are used consistently with the rest, to differentiate different arguments of the same type, which is a different purpose than you are suggesting here.

I don't really feel too strongly about any of the conventions in this case, since it's a bit of an outlier anyway. But I still think the most consistent approach would be t-last.

tsnobip:
OK, here is the version where everything should be as expected, with t-last, shadowing, etc.
I also versioned  Webapi__Dom.js, not sure if it was necessary or not, the versioning of JS files is kind of a mess since the latest versions of BS changed the output of pretty much all the JS files in the project.
Sorry again for the mistake.

tsnobip:
I mean I committed the Webapi__Dom.js, it includes changes directly caused by my PR like adding an export to Webapi__Dom__DomStringMap but also changes that are apparently caused by the use of a different version of BS.

glennsl:
Alright, that's no problem. Thank you!

PR #166, opened 2019-11-17
The 'features' argument (on Window.open_) should be optional

malthe:
(Because it is, according to the spec).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+* Changed `Window.open_` to have optional `features` argument
 
 ### 0.15.5 - (2019-10-09)
 * Added `keypress` event handler API
--- a/src/Webapi/Webapi__Dom/Webapi__Dom__Window.re
+++ b/src/Webapi/Webapi__Dom/Webapi__Dom__Window.re
@@ -77,7 +77,7 @@ module Impl = (T: {type t;}) => {
   [@bs.send.pipe : t_window] external matchMedia : string => mediaQueryList = ""; /* experimental, CSSOM View module */
   [@bs.send.pipe : t_window] external moveBy : (int, int) => unit = ""; /* experimental, CSSOM View module */
   [@bs.send.pipe : t_window] external moveTo : (int, int) => unit = ""; /* experimental, CSSOM View module */
-  [@bs.send.pipe : t_window] [@bs.return nullable] external open_ : (~url: string, ~name: string, ~features: string) => option(Dom.window) = "open"; /* yes, features is a stringly typed list of key value pairs, sigh */
+  [@bs.send.pipe : t_window] [@bs.return nullable] external open_ : (~url: string, ~name: string, ~features: string=?) => option(Dom.window) = "open"; /* yes, features is a stringly typed list of key value pairs, sigh */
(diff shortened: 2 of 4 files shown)


src/Webapi/Webapi__Dom/Webapi__Dom__Window.re
```diff
@@ -77,7 +77,7 @@ module Impl = (T: {type t;}) => {
   [@bs.send.pipe : t_window] external matchMedia : string => mediaQueryList = ""; /* experimental, CSSOM View module */
   [@bs.send.pipe : t_window] external moveBy : (int, int) => unit = ""; /* experimental, CSSOM View module */
   [@bs.send.pipe : t_window] external moveTo : (int, int) => unit = ""; /* experimental, CSSOM View module */
-  [@bs.send.pipe : t_window] [@bs.return nullable] external open_ : (~url: string, ~name: string, ~features: string) => option(Dom.window) = "open"; /* yes, features is a stringly typed list of key value pairs, sigh */
+  [@bs.send.pipe : t_window] [@bs.return nullable] external open_ : (~url: string, ~name: string, ~features: option(string)) => option(Dom.window) = "open"; /* yes, features is a stringly typed list of key value pairs, sigh */
```
glennsl: This isn't how you make arguments optional. I think the proper syntax is `~features: string=?`

tests/Webapi/Webapi__Dom/Webapi__Dom__Window__test.re
```diff
@@ -58,7 +58,7 @@ let _ = Window.getSelection(window);
 let _ = Window.matchMedia("max-height: 400", window);
 let _ = Window.moveBy(10, (-10), window);
 let _ = Window.moveTo(120, 300, window);
-let _ = Window.open_(~url="http://...", ~name="my window", ~features="menubar=yes", window);
+let _ = Window.open_(~url="http://...", ~name="my window", ~features=?Some("menubar=yes"), window);
```
glennsl: Then you don't need to make any change here, the argument can be passed like before but also left out.

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+* (Breaking) Changed `Window.open_` to have optional `features` argument
```
glennsl: It also won't be a breaking change.

glennsl:
Thanks! Could you also add another line in `Webapi__Dom__Window__test.re` for `Window.open_` with the argument left out, and check in the compiled js artifacts?

malthe:
@glennsl – I _think_ it's right:
```javascript
window.open("http://...", "my window", "menubar=yes");
window.open("http://...", "my window", undefined);
```
That is, the optional `features` argument becomes `undefined`, which is how it's supposed to be according to the [docs](https://bucklescript.github.io/docs/en/common-data-types#shared).

glennsl:
Perfect. Thank you!

> By the way, why are the JavaScript-files checked in – aren't they just build artifacts?

It's to be bale to inspect the generated code, to ensure it is as expected. And to be made aware of unexpected changes to the generated code from upgrading the compiler that might break existing code, or might require a change to behave like before.

PR #169, opened 2020-02-13
Rename Dom.Image to Dom.ImageData, add implementation for Dom.Image

TomiS:
Solves issue: https://github.com/reasonml-community/bs-webapi-incubator/issues/168

There is a backwards compatibility break in naming of the files.


TomiS:
Oh, I didn't realize there is a `HTMLImageElement` already. I guess proxy module would be a good idea. And I agree `ImageData` should probably be with canvas stuff. I'll see if I could fix the PR soon.

TomiS:
Will create a new PR from another branch.

PR #172, opened 2020-04-11
adding content-type

BlueHotDog:
Hey all, minor change, self described( i think)
https://developer.mozilla.org/en-US/docs/Web/API/Document/contentType

### DIFF
--- a/lib/js/tests/Webapi/Webapi__Dom/Webapi__Dom__Document__test.js
+++ b/lib/js/tests/Webapi/Webapi__Dom/Webapi__Dom__Document__test.js
@@ -10,6 +10,8 @@ var el = document.createElement("strong");
 
 document.characterSet;
 
+document.contentType;
+
 Webapi__Dom__Document.compatMode(document);
 
 document.doctype;
--- a/src/Webapi/Webapi__Dom/Webapi__Dom__Document.re
+++ b/src/Webapi/Webapi__Dom/Webapi__Dom__Document.re
@@ -18,6 +18,7 @@ module Impl = (T: {type t;}) => {
   Webapi__Dom__Node.nodeType(node) == Document ? Some(Obj.magic(node)) : None;
 
   [@bs.get] external characterSet : T.t => string = "";
+  [@bs.get] external contentType: T.t => string = "";
   [@bs.get] external compatMode : T.t => string /* compatMode enum */ = ""; /* experimental */
   let compatMode: T.t => Webapi__Dom__Types.compatMode =
     (self) => Webapi__Dom__Types.decodeCompatMode(compatMode(self));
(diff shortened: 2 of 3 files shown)


src/Webapi/Webapi__Dom/Webapi__Dom__HtmlDocument.re
```diff
@@ -3,11 +3,13 @@ module Impl = (T: {type t;}) => {
 
   [@bs.get] [@bs.return nullable] external activeElement : t_htmlDocument => option(Dom.element) = "";
   [@bs.get] [@bs.return nullable] external body : t_htmlDocument => option(Dom.element) = ""; /* returns option HTMLBodyElement */
+  [@bs.get] external characterSet : t_htmlDocument => string = "";
```
glennsl, line 6: Why has this been moved here from `Document`? The documentation you refer to lists it as a property of `Document`, and `as HtmlDocument` inherits everything from `Document` it should be available here anyway.

src/Webapi/Webapi__Dom/Webapi__Dom__HtmlDocument.re
```diff
@@ -3,11 +3,13 @@ module Impl = (T: {type t;}) => {
...
   [@bs.set] external setBody : (t_htmlDocument, Dom.element) => unit = "body"; /* accepth HTMLBodyElement */
   [@bs.get] external cookie : t_htmlDocument => string = "";
   [@bs.set] external setCookie : (t_htmlDocument, string) => unit = "cookie";
   [@bs.get] [@bs.return nullable] external defaultView : t_htmlDocument => option(Dom.window) = "";
   [@bs.get] external designMode : t_htmlDocument => string /* designMode enum */ = "";
+  [@bs.get] external contentType: T.t => string = "";
```
glennsl, line 12: This is marked as experimental on MDN and seems to be unavailable on IE11 and Opera Mini. I'm hesitant to include APIs that aren't fully standardized, although this seems pretty established. It should at least have a comment noting that it's experimental here as well. It should also be in `Document`, not `HtmlDocument`.

src/Webapi/Webapi__Dom/Webapi__Dom__HtmlDocument.re
BlueHotDog: weird, didnt find it. but will move it here.
In general i wonder what's the actal distinction between document and htmldocument.
In the docs it means that for all porpuses its the same thing when dealing with the web.

glennsl: You have also removed it from `Document`, and from its test, so it is actually moved, not added. Seems like it took you a few iterations to get here though, so I guess a mixup happened somewhere.

glennsl: `HTMLDOcument` is a subtype of `Document`, ie. more specific. In addition to HTML Documents a `Document` can also be an `XMLDocument` and I think also an `SVGDocument`. These kinds of documents also exist on the web.

PR #175, opened 2020-04-14
WIP: Add support for attaching beforeunload event listeners

TomiS:
The intention is to enable using this browser feature:
https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload

There seems to already be a type for [`Webapi.BeforeUnloadEvent](https://github.com/reasonml-community/bs-webapi-incubator/blob/master/src/Webapi/Webapi__Dom/Webapi__Dom__BeforeUnloadEvent.re), but not a way to bind to it. So this PR adds that.

TODO: 
Bind deleting a field from object somehow to enable equivalent of this
```
delete e['returnValue'];
```


TomiS:
I'll also close this. And create a new one.

PR #176, opened 2020-04-14
Implement adding event listeners for beforeunload

TomiS:


src/Webapi/Webapi__Dom/Webapi__Dom__BeforeUnloadEvent.re
```diff
@@ -6,3 +6,8 @@ include Webapi__Dom__Event.Impl({ type nonrec t = t; });
 [@bs.new] external makeWithOptions : (string, Js.t({..})) => t = "BeforeUnloadEvent";
 
 [@bs.get] external returnValue : t => string = "";
+/**
+ * returnValue cannot be unset because BS cannot (afaik) bind `delete e['returnValue'];`
+ * Setting the value to undefined does not cause expected behavior.
+ */
+[@bs.set] external setReturnValue: (t, string) => unit = "returnValue";
```
glennsl, line 13: `return`Value` si actually defined on `Event`, which this inherits from. Could you move it there instead?

It;s also possible to write a function that does a `delete`, but you'll have to use `%raw`. Something like this should work I think:

```reason
let deleteReturnValue: t => unit = [%raw event => {|
  delete event['returnValue']
|}];
```

TomiS: Ah, indeed it is defined in [Event](https://developer.mozilla.org/en-US/docs/Web/API/Event/returnValue). I'll move it there and try adding deleting.

TomiS:
Ok, finally managed to do some changes. Is this better?

tests/Webapi/Dom/Webapi__Dom__Event__test.re
```diff
@@ -14,7 +14,10 @@ let _ = target(event);
 let _ = timeStamp(event);
 let _ = type_(event);
 let _ = isTrusted(event);
+let _ = returnValue(event);
 
 preventDefault(event);
 stopImmediatePropagation(event);
 stopPropagation(event);
+setReturnValue(event, "Test");
+deleteReturnValue(event);
```
glennsl, line 23: Can you commit the JS artifacts as well so we can ensure this generates correct code?

glennsl (commented):
Hmm, yeah that's a problem. The [DOM specification for `Event`](https://dom.spec.whatwg.org/#interface-event) considers `returnValue`"historical", whatever that means, so perhaps we could get away with having the setter only accept a string, and to have it only on `BeforeUnloadEvent`.

But either way the getter can return both boolean and string values, so to be safe it needs to account for that. Not sure if we do that anywhere here already, but if we do this ought to mimic that. Otherwise I think using `Js.Types.classify` and returning a polymorphic variant is an OK approach.

PR #182, opened 2020-05-20
Downcast DOM types by using `instanceof` to check types

yawaramin:
This should be more robust than checking some properties of the objects themselves. Since the Web API classes like `HTMLDocument`, `HTMLElement` et al. are globally available in the browser we can just declare their constructors using `[@bs.val]` without having to worry about importing them.

### DIFF
--- a/src/Webapi/Webapi__Dom/Webapi__Dom__Document.re
+++ b/src/Webapi/Webapi__Dom/Webapi__Dom__Document.re
@@ -1,17 +1,13 @@
 module Impl = (T: {type t;}) => {
   external asDocument : T.t => Dom.document = "%identity";
 
-  let asHtmlDocument: T.t => Js.null(Dom.htmlDocument) = [%bs.raw
-    {|
-    function (document) {
-      return document.doctype.name === "html" ?  document : null;
+  let asHtmlDocument: T.t => option(Dom.htmlDocument) = [%raw {|
+    function(document) {
+      if (document instanceof HTMLDocument) return document;
     }
-  |}
-  ];
-  [@deprecated "Will fail if no doctype is defined, consider using unsafeAsHtmlDocument instead"]
-  let asHtmlDocument: T.t => option(Dom.htmlDocument) =
-    (self) => Js.Null.toOption(asHtmlDocument(self));
+  |}];
 
+  [@deprecated "Unsafe cast, use [ashtmlDocument] instead"]
   external unsafeAsHtmlDocument : T.t => Dom.htmlDocument = "%identity";
 
   let ofNode = (node: Dom.node) : option(T.t) =>
--- a/src/Webapi/Webapi__Dom/Webapi__Dom__Element.re
+++ b/src/Webapi/Webapi__Dom/Webapi__Dom__Element.re
@@ -5,18 +5,13 @@ let ofNode = (node: Dom.node) : option('a) =>
     None;
 
 module Impl = (T: {type t;}) => {
-  let asHtmlElement: T.t => Js.null(Dom.htmlElement) = [%bs.raw
-    {|
-    function (element) {
-      // BEWARE: Assumes "contentEditable" uniquely identifies an HTMLELement
-      return element.contentEditable !== undefined ?  element : null;
+  let asHtmlElement: T.t => option(Dom.htmlElement) = [%raw {|
+    function(element) {
+      if (element instanceof HTMLElement) return element;
     }
-  |}
-  ];
-  [@deprecated "asHtmlElement uses a weak heuristic, consider using unsafeAsHtmlElement instead"]
-  let asHtmlElement: T.t => option(Dom.htmlElement) =
-    (self) => Js.Null.toOption(asHtmlElement(self));
+  |}];
 
+  [@deprecated "Unsafe cast, use [asHtmlElement] instead"]
   external unsafeAsHtmlElement : T.t => Dom.htmlElement = "%identity";
   let ofNode: Dom.node => option(T.t) = ofNode;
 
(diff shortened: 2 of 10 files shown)


src/Webapi/Webapi__Dom/Webapi__Dom__Element.re
```diff
@@ -5,18 +5,13 @@ let ofNode = (node: Dom.node) : option('a) =>
...
 
 module Impl = (T: {type t;}) => {
-  let asHtmlElement: T.t => Js.null(Dom.htmlElement) = [%bs.raw
-    {|
-    function (element) {
-      // BEWARE: Assumes "contentEditable" uniquely identifies an HTMLELement
-      return element.contentEditable !== undefined ?  element : null;
+  let asHtmlElement: T.t => option(Dom.htmlElement) = [%raw {|
+    function(element) {
+      if (element instanceof HTMLElement) return element;
```
TheSpyder, line 10: This won't work for elements created in a different window (e.g. an iframe). The old code wasn't great, but it at least passed that case. Here's how we do it in TinyMCE, using `isPrototypeOf` instead of `instanceof`:
https://github.com/tinymce/tinymce/blob/develop/modules/sand/src/main/ts/ephox/sand/api/HTMLElement.ts#L15-L20

src/Webapi/Webapi__Dom/Webapi__Dom__Element.re
```diff
@@ -5,18 +5,13 @@ let ofNode = (node: Dom.node) : option('a) =>
...
     }
-  |}
-  ];
-  [@deprecated "asHtmlElement uses a weak heuristic, consider using unsafeAsHtmlElement instead"]
-  let asHtmlElement: T.t => option(Dom.htmlElement) =
-    (self) => Js.Null.toOption(asHtmlElement(self));
+  |}];
 
+  [@deprecated "Unsafe cast, use [asHtmlElement] instead"]
```
TheSpyder, line 14: This deprecation warning is a bit onerous for library authors building on top of `bs-webapi`, I'm going to have to add a non-deprecated replacement and further diverge from using `bs-webapi` directly.

I have 11 places in my code where I know with certainty that I'm in a HTML tree structure, but instead of lightweight code:
`switch(elm->Element.parentElement->Element.unsafeAsHtmlElement)`
which generates
```
    var parent = elm$1.parentElement;
    if (parent == null) {
      // no parent code
    }
```
I now have to do:
`switch(elm->HtmlElement.parentElement->Belt.Option.flatMap(Element.asHtmlElement))`
which generates 
```
    var parent = Belt_Option.flatMap(Caml_option.nullable_to_opt(elm.parentElement), Webapi__Dom__Element.asHtmlElement);
    if (parent === undefined) {
      // no parent code
    }
```
(it could be shrunk with nested switch statements instead of flatMap, but that's unacceptable imo)

This difference is partly due to how `nullable` externals are generated. When used immediately they are handled with optimal JS. When passed as values, however, they have to be wrapped in a `nullable_to_opt` function call to normalise them with regular option values.

yawaramin: Yeah, you're right. The function names are already marked as `unsafe`. I'll publish a new version and un-deprecate them.

src/Webapi/Webapi__Dom/Webapi__Dom__Element.re
yawaramin: Fixed in https://github.com/reasonml-community/bs-webapi-incubator/blob/4f9a3624f47571d0474b2e2e6de5f7ab0bea340a/src/Webapi/Webapi__Dom/Webapi__Dom__Element.re#L8 , thanks for pointing this out!

yawaramin: Thanks for pointing this out, should be fixed in https://github.com/reasonml-community/bs-webapi-incubator/blob/08e00d32c280ad6de8209114e825a56f842cbaa3/src/Webapi/Webapi__Dom/Webapi__Dom__Element.re#L9

TheSpyder: Cool. That's pretty much how the TinyMCE code works, thanks to the `resolve` helper function, I can't remember why we stuck to `isPrototypeOf` instead of `instanceof` for the final check but it should be the same.

src/Webapi/Webapi__Dom/Webapi__Dom__Element.re
TheSpyder: Thanks!

src/Webapi/Webapi__Dom/Webapi__Dom__Document.re
```diff
@@ -1,17 +1,13 @@
 module Impl = (T: {type t;}) => {
   external asDocument : T.t => Dom.document = "%identity";
 
-  let asHtmlDocument: T.t => Js.null(Dom.htmlDocument) = [%bs.raw
-    {|
-    function (document) {
-      return document.doctype.name === "html" ?  document : null;
+  let asHtmlDocument: T.t => option(Dom.htmlDocument) = [%raw {|
+    function(document) {
+      if (document instanceof HTMLDocument) return document;
```
TheSpyder, line 6: @yawaramin I forgot to mention, and check for myself, but this has the same problem as `instanceof HTMLElement`. Now that I've merged my upgrade to the new release, I've hit this bug.

Perhaps it would be better to create a re-usable way to find the instance so you're not adding a bunch of `null` checks here as well 🤔

yawaramin: Whoops, thanks again. Fixed and published 0.16.2. For the moment I'm not yet sure how best to generalize the check, so just brute-forcing it for now. Would be really cool if everybody supported optional coalescing already!

TheSpyder: I don't have to support IE11 so I could 😂

I'll work on a PR soon that combines them. It shouldn't be hard.

PR #184, opened 2020-05-26
Fix DomStringMap.unsafeDeleteKey to compile with BuckleScript 8

johnridesabike:
BuckleScript 8 (currently `bs-platform@dev`) removes the `[%raw (key, map) => "delete map[key];"]` syntax. (See https://github.com/BuckleScript/bucklescript/issues/4415) This PR changes that line of code so it can compile when BS8 releases.

### DIFF
--- a/lib/js/tests/Webapi/Webapi__Dom/Webapi__Dom__DomStringMap__test.js
+++ b/lib/js/tests/Webapi/Webapi__Dom/Webapi__Dom__DomStringMap__test.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var Curry = require("bs-platform/lib/js/curry.js");
 var Webapi = require("../../../src/Webapi.js");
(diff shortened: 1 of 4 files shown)


lib/js/tests/Webapi/Webapi__Dom/Webapi__Dom__DomStringMap__test.js
```diff
@@ -10,7 +11,7 @@ Webapi__Dom__DomStringMap.set("fooKey", "barValue", dataset);
 
 Webapi__Dom__DomStringMap.get("fooKey", dataset);
 
-Webapi__Dom__DomStringMap.unsafeDeleteKey("fooKey", dataset);
+Curry._2(Webapi__Dom__DomStringMap.unsafeDeleteKey, "fooKey", dataset);
```
johnridesabike, line 14: Note that if you compile with the latest BuckleScript, this does not have runtime currying.

PR #186, opened 2020-07-01
Namespace EventPhase `None` constructor in its own module

### LINKED ISSUE #84: Opening Webapi.Dom shadows `None`
Reason beginner here, so apologies in advance if this is not a bug.

If I `open Webapi.Dom`, it shadows the `None` constructor, and I cannot have optional variables in the same file.

It currently shadows None with `Webapi.Dom.eventPhase`.

yawaramin:
Ping @TheSpyder

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 
+### 0.17.1 (2020-07-01)
+* Deprecated `Webapi.Dom.eventPhase` in favour of
+  `Webapi.Dom.EventPhase.t`
+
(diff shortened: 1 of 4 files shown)


TheSpyder:
Interesting, this is more or less option 2 as discussed back in September. Spocke and I are in favour of it, but it looks like we never responded to the question about why. Interesting that it's your first reaction as well :)

yawaramin:
Yeah more or less. I think it sidesteps one of the issues Glenn originally pointed out, which is moving the `eventPhase` type into the `Dom.Event` module.

PR #189, opened 2020-08-04
Add labeled argument to Url.makeWithBase.

johnridesabike:
This is a minor (but breaking) change that adds a `~base` label to the second argument of `Url.makeWithBase`. I think the current signature, `(string, string) => t`, is ambiguous as to which string is the “base."

### DIFF
--- a/src/Webapi/Webapi__Url.re
+++ b/src/Webapi/Webapi__Url.re
@@ -21,7 +21,7 @@ module URLSearchParams = {
 type t;
 
 [@bs.new] external make: string => t = "URL";
-[@bs.new] external makeWithBase: (string, string) => t = "URL";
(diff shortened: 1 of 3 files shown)


src/Webapi/Webapi__Url.re
```diff
@@ -21,7 +21,7 @@ module URLSearchParams = {
 type t;
 
 [@bs.new] external make: string => t = "URL";
-[@bs.new] external makeWithBase: (string, string) => t = "URL";
+[@bs.new] external makeWithBase: (string, ~base: string) => t = "URL";
```
yawaramin, line 24: Since this is a breaking change, I think it's better to make a new function and deprecate this one. The new one can be `makeWith(string, ~base)`.

johnridesabike: Done.

PR #191, opened 2020-08-04
Fixes incorrect array_like type for URLSearchParams.entries()

### LINKED ISSUE #190: Binding for URLSearchParams.entries() is incorrect
https://github.com/reasonml-community/bs-webapi-incubator/blob/master/src/Webapi/Webapi__Url.re#L9

The binding says a return type of `Js.Array.array_like(string)`, but should be `Js.Array.array_like((string, string))`

https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

illusionalsagacity:
I believe I followed the contributing guide, let me know if you would like any other changes.

This PR is to change the return value of the `Webapi__Url.URLSearchParams.entries` binding from `array_like(string)` to `array_like((string, string))`

```sh
$ node lib/js/tests/Webapi/Webapi__Url__test.js
value1 key1
value2 key2
[ [ 'key1', 'value1' ], [ 'key2', 'value2' ] ]
```

Fixes #190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ### 0.18.1
 
-* Fixed the binding for `Webapi.Url.URLSearchParams.entries` to return
-  `(string, string)`
+* Fixed the binding for `Webapi.Url.URLSearchParams.entries` to return `(string, string)`
 * Added `Webapi.Dom.Node.replaceChild`
 
 ### 0.18.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
-### 0.18.1
+### 0.19.0
 
 * Fixed the binding for `Webapi.Url.URLSearchParams.entries` to return `(string, string)`
+
+### 0.18.1
+
 * Added `Webapi.Dom.Node.replaceChild`
 
 ### 0.18.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,6 @@
 ### 0.19.0
 
 * Fixed the binding for `Webapi.Url.URLSearchParams.entries` to return `(string, string)`
-
-### 0.18.1
-
 * Added `Webapi.Dom.Node.replaceChild`
 
 ### 0.18.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ### 0.18.1
 
+* Fixed the binding for `Webapi.Url.URLSearchParams.entries` to return
+  `(string, string)`
 * Added `Webapi.Dom.Node.replaceChild`
 
 ### 0.18.0
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bs-webapi",
-  "version": "0.18.1",
+  "version": "0.19.0",
   "description": "Reason + BuckleScript bindings to DOM",
   "repository": {
     "type": "git",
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+### 0.18.1
+
+* Fixed the binding for `Webapi.Url.URLSearchParams.entries` to return
+  `(string, string)`
```
yawaramin, line 4: Could you join this with the previous line?

tests/Webapi/Webapi__Url__test.re
```diff
@@ -1,4 +1,6 @@
 open Webapi.Url;
 
 let params = URLSearchParams.make("key1=value1&key2=value2");
-URLSearchParams.forEach(Js.log2, params);
\ No newline at end of file
+URLSearchParams.forEach(Js.log2, params);
+let entries: array((string, string)) = Js.Array.from(URLSearchParams.entries(params));
```
yawaramin, line 5: Could you remove the type annotation? In keeping with the othe variables in test files. Thanks.

illusionalsagacity: So for this one I get this error:

```
Error: The type of this expression, array('_weak1),
       contains type variables that cannot be generalized
```

When I look at the `js_array.ml` implementation in bucklescript, it's `'a array_like -> 'b array`, so the compiler doesn't like it.

illusionalsagacity:
@yawaramin would you prefer rebasing away the merge conflicts or just a merge?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bs-webapi",
-  "version": "0.18.0",
+  "version": "0.18.1",
```
glennsl, line 3: This is a breaking change, so it should be 0.19

tests/Webapi/Webapi__Url__test.re
glennsl: Ouch, well that's clearly wrong. Good opportunity for contributing a fix to bucklescript though!

yawaramin: @illusionalsagacity ah, right, value restriction. What I'm doing nowadays is writing 'tester' functions like this:

```reason
let test_entries = params => params |> URLSearchParams.entries |> Js.Array.from;
```

Since it's inside a function, it shouldn't hit the value restriction.

package.json
yawaramin: Agreed, @illusionalsagacity would you mind bumping to 0.19?

yawaramin:
@illusionalsagacity re: merge or rebase, whichever you prefer. At the end I will squash and merge anyway.

tests/Webapi/Webapi__Url__test.re
yawaramin: Btw the wrong signature is a known issue https://github.com/BuckleScript/bucklescript/issues/3272

illusionalsagacity:
Alright that should resolve those

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### 0.19.0
+
+* Fixed the binding for `Webapi.Url.URLSearchParams.entries` to return `(string, string)`
+
 ### 0.18.1
 
 * Added `Webapi.Dom.Node.replaceChild`
```
yawaramin, line 7: Can you fold this log entry into 0.19.0, and get rid of the 0.18.1 release here, thx

PR #192, opened 2020-08-04
Implement binding for Node.replaceChild

### LINKED ISSUE #179: Adding replaceChild
Add missing replaceChild for `Webapi.Dom.Node`

illusionalsagacity:
https://developer.mozilla.org/en-US/docs/Web/API/Node/replaceChild

The `replaceChild` api doesn't seem to care if the two arguments are the same type, so I bound it as returning the same type as the old child.

Fix #179


yawaramin:
Thanks!

PR #194, opened 2020-08-16
Add bindings for `Navigator`

forrest-akin:


yawaramin:
Thanks! Will review soon.

src/Webapi/Dom/Webapi__Dom__Window.re
```diff
@@ -4,7 +4,6 @@ type frameList; /* array-like, WindowProxy? */
 type idleDeadline; /* Cooperative Scheduling of Background Tasks */
 type locationbar; /* "bar object" */
 type menubar; /* "bar object" */
-type navigator;
```
yawaramin, line 7: Deleting this type will break backward-compatibility, so instead can you alias it to the newly-introduced `Webapi__Dom__Navigator.t` type and deprecate the alias, thx

src/Webapi/Dom/Webapi__Dom__Navigator.re
```diff
@@ -0,0 +1,66 @@
...
+    | `Float32Array(Float32Array.t)
+    | `Float64Array(Float64Array.t)
+    | `DataView(DataView.t)
+    | `Blob(Webapi__Blob.t)
+    | `FormData(Fetch.FormData.t)
+    | `URLSearchParams(Webapi__Url.URLSearchParams.t)
+    | `String(string)
+  ]) => bool = "";
+[@bs.send.pipe : t] external share : shareOptions => Js.Promise.t(unit) = "";
+[@bs.send.pipe : t] external vibrate : (
```
yawaramin, line 62: Since there are only two types of allowed parameters here, can you use two different bindings i.e. `vibrate` and `vibrateArray`? That's the typical style used in `bs-webapi`, and I recommended the `@bs.unwrap` style for `sendBeacon` because it takes so many different types that it would have exploded the number of bindings.

src/Webapi/Dom/Webapi__Dom__Navigator.re
```diff
@@ -0,0 +1,66 @@
...
+  [@bs.unwrap] [
+    | `Int(int)
+    | `IntList(list(int))
```
yawaramin, line 65: For `vibrateArray` we need to ensure that the parameter type is `array(int)`, because a Reason array corresponds to a JavaScript array. It can't be a `list`.

src/Webapi/Webapi__Dom.re
```diff
@@ -71,6 +72,7 @@ include Webapi__Dom__Types;
 [@bs.val] external document : Dom.document = "document";
 [@bs.val] [@bs.scope "window"] external history : Dom.history = "history";
 [@bs.val] [@bs.scope "window"] external location : Dom.location = "location";
+[@bs.val] [@bs.scope "window"] external navigator : Webapi__Dom__Navigator.t = "navigator";
```
yawaramin, line 75: I realize there are similar bindings above it, but I really don't think `Webapi.Dom.navigator` is a sensible place to access a `Navigator` object from. Since the `Webapi.Dom.Window.navigator` binding already exists, we don't need another one here.

tests/Webapi/Dom/Webapi__Dom__Navigator__test.re
```diff
@@ -0,0 +1,28 @@
+open Webapi.Dom;
```
yawaramin, line 1: Thanks for writing these tests. I know it's a bit annoying, but can you also check in the compiled test output? You'll need to do:

    git add --force lib/js/tests/Webapi/Dom/Webapi__Dom__Navigator__test.js

forrest-akin:
@yawaramin Thanks for pointing me to the CONTRIBUTION page; missed it on my first pass. I think everything should be in order now. Per [Can I Use](https://caniuse.com/#search=navigator%20api), I omitted properties/methods that are deprecated/obsolete, commented out anything experimental, so all bindings in the module should be stable. lmk if I missed anything!

PR #195, opened 2020-08-18
feat: add bindings for HTMLFormControlsCollection, RadioNodeList, etc

illusionalsagacity:
Needs to wait for https://github.com/rescript-lang/rescript-compiler/pull/4622 at the moment, so the 0.20 version is just placeholder

Not really sure how to proceed with the `namedItem` binding, when I tried to put `isRadioNodeList` into the `Webapi__Dom__RadioNodeList` module, it gave me an error that the `'a` couldn't be generalized or something like that.


yawaramin:
Thanks, I'll try to review soon. Just a note, that after the rescript-compiler PR merge, it will be published in BS 8.something). The problem is, that would require bs-webapi to upgrade to 8.something to be able to use the new types introduced there. And BS 8+ has some backward-incompatible changes, which means upgrading to that version would prevent some folks from being able to upgrade bs-webapi. In the interim, I would recommend to introduce the types independently here and then later we can decide to alias them to the BS types for backward-compatibility.

illusionalsagacity:
> Thanks, I'll try to review soon. Just a note, that after the rescript-compiler PR merge, it will be published in BS 8.something). The problem is, that would require bs-webapi to upgrade to 8.something to be able to use the new types introduced there. And BS 8+ has some backward-incompatible changes, which means upgrading to that version would prevent some folks from being able to upgrade bs-webapi. In the interim, I would recommend to introduce the types independently here and then later we can decide to alias them to the BS types for backward-compatibility.

Yeah I was thinking of that as well, just have the version to get it linked against my local rescript version for now.

src/Webapi/Dom/Webapi__Dom__Document.re
```diff
@@ -34,6 +34,8 @@ module Impl = (T: {type t;}) => {
   [@bs.get] external implementation : T.t => Dom.domImplementation = "";
   [@bs.get] external lastStyleSheetSet : T.t => string = "";
   [@bs.get] [@bs.return nullable] external pointerLockElement : T.t => option(Dom.element) = ""; /* experimental */
+  /** @since 0.20.0 */
+  [@bs.get] external forms : T.t => Dom.htmlCollection = "";
```
illusionalsagacity, line 38: would be nice if we could type as `Dom.htmlCollection(Dom.htmlFormElement)`

PR #196, opened 2020-08-20
feat: HtmlVideoElement, HtmlMediaElement, TextTrack, VideoTrack

### LINKED ISSUE #173: adding bindings to HtmlVideoElement
Hi, I want to add bindings to HtmlVideoElement(defined here: https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement). seems like there's a missing bindings for HtmlMediaElement(also the type is missing on Ocaml Dom).
Should i take a stab at it?
If so, how would i do that without the relevant ocaml type?

illusionalsagacity:
Got a couple of questions here;

Should we append `Exn` for methods that potentially throw? Or maybe we catch it for convenience and return a result? 

Along the same lines, I put in some variants to hold the enums in some (but not all) cases. I kinda like this, as it makes life easier when consuming the bindings. But, willing to remove entirely or cover the other ones. The main problem I ran into was I couldn't type a return value as:

```
external foo: unit => [@bs.string] [`one | `two | `zero] = "";
```

So hiding the binding and then wrapping is necessary.

fixes #173

### DIFF
--- a/src/Webapi/Dom/Webapi__Dom__HtmlMediaElement.re
+++ b/src/Webapi/Dom/Webapi__Dom__HtmlMediaElement.re
@@ -1,0 +1,139 @@
+/**
+ * Spec: https://html.spec.whatwg.org/multipage/media.html#htmlmediaelement
+ */
+module Impl = (T: {type t;}) => {
+  type t_htmlMediaElement = T.t;
+
+  type mediaProvider; // TODO: one of MediaStream, MediaSource, or Blob
+
+  /** Properties */
+  [@bs.get] external buffered: t_htmlMediaElement => Webapi__TimeRanges.t = "buffered";
+
+  // This is part of the recommendation spec
+  // type mediaController; // TODO
+  // [@bs.get] [@bs.return nullable] external controller: t_htmlMediaElement => option(mediaController) = "controller";
+  // [@bs.set] external setController: (t_htmlMediaElement, mediaController) => unit = "controller";
+
+  [@bs.get] external controls: t_htmlMediaElement => bool = "controls";
+  [@bs.set] external setControls: (t_htmlMediaElement, bool) => unit = "controls";
+
+  /** One or more of "nodownload", "nofullscreen", and "noremoteplayback" */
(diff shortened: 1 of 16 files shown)


src/Webapi/Dom/Webapi__Dom__HtmlMediaElement.re
```diff
@@ -0,0 +1,139 @@
+/**
+ * Spec: https://html.spec.whatwg.org/multipage/media.html#htmlmediaelement
+ */
+module Impl = (T: {type t;}) => {
+  type t_htmlMediaElement = T.t;
+
+  type mediaProvider; // TODO: one of MediaStream, MediaSource, or Blob
```
illusionalsagacity, line 7: Thinking this might be a `Webapi__MediaProvider.t` instead of a local phantom type

src/Webapi/Dom/Webapi__Dom__HtmlMediaElement.re
```diff
@@ -0,0 +1,139 @@
...
+  /** Methods */
+
+  [@bs.send] external addTextTrack: (t_htmlMediaElement, Webapi__TextTrack.textTrackKind, ~label: string, ~language: string, unit) => Webapi__TextTrack.t = "addTextTrack";
+
+  // return is one of "probably", "maybe", or ""
+  [@bs.send] external canPlayType: (t_htmlMediaElement, string) => string = "canPlayType";
+  // [@bs.send] external fastSeek: (t_htmlMediaElement, float) => unit = "fastSeek"; // experimental
+  [@bs.send] external load: t_htmlMediaElement => unit = "load";
+  [@bs.send] external pause: t_htmlMediaElement => unit = "pause";
+  [@bs.send] external play: t_htmlMediaElement => Js.promise(unit, string /* [`NotAllowedError | `NotSupportedError] */) = "play";
```
kazcw, line 128: This is the deprecated promise type. The rest of this library is using `Js.Promise.t`

illusionalsagacity: Thanks!

